PR #68, opened 2026-06-11
Feat/node entrance animation

nazarli-shabnam:


nazarli-shabnam:
### Code review

Found 1 issue:

1. **OG image file is never referenced by the share URL's HTML** — `opengraph-image.tsx` is placed alongside `route.ts` at `app/api/share/[id]/`. Next.js injects the generated `<meta property="og:image">` tag into the HTML of the route it lives in (`/api/share/[id]`), but that route returns JSON (not HTML), so there is no HTML `<head>` to inject into. Meanwhile, the URL users actually share is `/canvas?share=<id>` (see `buildShareUrl` in `hooks/useCanvasPersistence.ts`), and the canvas page has no `generateMetadata` or dynamic OG image pointing at this endpoint. The image PNG is generated and served correctly, but crawlers (Twitter, OG debuggers) will never discover it.

  Fix: either move the share page to a proper page route (e.g. `app/share/[id]/page.tsx` + `app/share/[id]/opengraph-image.tsx`) and update `buildShareUrl` to point there, or add a `generateMetadata` to `app/canvas/page.tsx` that reads the `?share=` param and sets `openGraph.images` to `/api/share/${shareId}/opengraph-image`.

  https://github.com/Nazm-AI/open-hikmah/blob/94ebce5ce51834b4a3697ac58935b9e7d3bdc27d/app/api/share/%5Bid%5D/opengraph-image.tsx#L1-L57

PR #139, opened 2026-07-04
chore: add prettier format check

hfl0506:
## Summary

- Added Prettier config and ignore file.
- Added `format` and `format:check` scripts.
- Added Prettier to lint-staged.
- Added `format:check` to CI.
- Ran baseline formatting across the repo.
- Updated `bun.lock` with Prettier.

## Related Issues
Resolves #134 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

- [x] No AI or theological changes in this PR
- [ ] Claude prompt modified — described below
- [ ] New/updated divine name descriptions — verified against Maturidi/Hanafi sources
- [ ] Changes to theological framing — described below

**Details** (if applicable):

## Testing

- [x] `bun run format:check` passes locally
- [x] `bun run lint` passes locally
- [x] `bun run typecheck` passes locally
- [x] `fnm exec --using=20 sh .husky/pre-commit` passes locally
- [ ] New tests added for new functionality
- [ ] Tested in browser (for UI changes)

## Checklist

- [ ] Branch is up to date with `main`
- [x] Conventional commits
- [x] No secrets or real API keys in the diff
- [ ] `.env.example` updated if new environment variables were added
- [ ] `README.md` updated if the feature changes the public

Issue #137, opened 2026-07-03
Tracking: Add automated PR gates (e2e, formatting, a11y, bundle-size) before human review


nazarli-shabnam:
## Goal
Every PR should pass a set of automated checks before it reaches human review, catching regressions mechanically instead of relying on a reviewer to notice them. Currently CI (`.github/workflows/ci.yml`) only runs lint, typecheck, unit tests, integration tests, and build — no e2e, formatting, accessibility, or bundle-size checks exist.

## Sub-issues
- [x] #133 — Add Playwright E2E test suite + CI job
- [x] #134 — Add Prettier formatting check to CI
- [x] #135 — Add automated accessibility (axe) checks to e2e suite (depends on #133)
- [x] #136 — Add bundle size / build budget check to CI

## Wiring these in as an actual gate (not just visibility)
Once the checks above exist, two more steps are needed for them to actually *block* merge/review rather than just report:

1. **Branch protection**: mark the new CI jobs (`e2e`, `format:check`, bundle-size check) as **required status checks** for `main` in GitHub repo Settings → Branches. This is a repo-settings change, not something committed in-repo — easy to forget after adding a check, so calling it out explicitly here.
2. **PR template**: update `.github/PULL_REQUEST_TEMPLATE.md`'s Testing checklist — replace the manual "Tested in browser" item with a note that e2e/a11y/bundle-size checks now run automatically in CI, so reviewers know not to re-verify those by hand.

## Suggested order
1 → 2 → 3 (in parallel with 2) → 4, then branch-protection + PR-template wiring last, once all four checks are green on `main`.

PR #145, opened 2026-07-04
chore: add bundle size check in ci

### LINKED ISSUE #136: Add bundle size / build budget check to CI
## Problem
There is no bundle-size or build-budget tracking in the project. A dependency or code change that significantly bloats the client JS bundle would currently ship without any automated warning.

## Proposed fix
- Add `size-limit` (or `bundlewatch`) as a dev dependency, configured against the Next.js production build output (`.next/`).
- Set an initial budget based on the current build's actual bundle size plus reasonable headroom (avoid picking an arbitrary number that immediately fails on `main`).
- Add a CI step to (or a new job depending on) the existing `build` job in `.github/workflows/ci.yml` that runs the size check after `bun run build`, failing the PR if the budget is exceeded and ideally commenting the size diff on the PR (most of these tools support a GitHub Action integration for PR comments).

## Why this matters
Bundle-size regressions are easy to introduce accidentally (a new heavy dependency, an un-code-split import) and hard for a human reviewer to catch just by reading a diff — this is exactly the kind of check better delegated to automation than review.

hfl0506:
## Summary

- Added `size-limit` budget checking for Next.js production client JS output.
- Set the initial bundle budget from the current build size with ~25% headroom.
- Added the size check to CI after `bun run build` so bundle regressions fail PRs.

## Related Issue
Resolve #136 

## Type of change

- [x] New feature
- [x] Refactor / chore

## AI / Theological changes

**Details** (if applicable):

## Testing

- [x] `bun install --frozen-lockfile` passes locally
- [x] `bun run build` passes locally
- [x] `bun run size` passes locally

## Checklist

- [ ] Branch is up to date with `main`
- [ ] Conventional commits
- [x] `.env.example` updated if new environment variables were added


hfl0506:
Hi @nazarli-shabnam I use 25% of bundle size to be headroom, let me know if this is make sense to you. Thank you

nazarli-shabnam (changes requested):
Thanks for this, @hfl0506 — bundle-size budgets are exactly what #136 asked for, and the setup (size-limit + a CI step right after `bun run build`) is the right shape for it.

Two things before this is mergeable:

1. **Merge conflict**: this branch conflicts with `main` in `bun.lock` — both this PR and the just-merged Prettier PR (#139) added a devDependency entry at the same alphabetical spot. Could you merge/rebase `main` into your branch and resolve it? Should just be keeping both the `prettier` and `size-limit` lines (and re-running `bun install` to regenerate the lockfile cleanly).
2. CodeRabbit's two nitpicks are worth a look but not blocking:
   - No PR comment with the size delta yet — fine as a follow-up rather than scope creep here.
   - The `.next/static/**/*.js` glob budgets *all* shipped client JS (every chunk), not just the initial-load bundle. Reasonable choice, just flagging so it's a conscious one.

Once the conflict's resolved this looks good to me — requesting changes only for that, not the approach.

nazarli-shabnam:
Hey @hfl0506 — yeah, that makes sense to me. 

The point of the buffer is that the check shouldn't fail on every small, expected bundle growth (a new dependency, a new route) but should still catch a real regression (e.g. someone accidentally bundling a large library or duplicating a chunk). 25% is a sensible starting point for a first pass — we can tighten it later once we've seen how the number moves across a few PRs.

No changes needed there, good call.

hfl0506:
Hi @nazarli-shabnam Update change, please help me take a look when you have a moment. Thanks

nazarli-shabnam (changes requested):
Did a careful manual pass on this (CodeRabbit's review wasn't catching the real issue here). The conflict resolution looks good, and the "Guard the bundle report against missing JSON" fix is solid. But there's one real bug in the new PR-comment feature that I verified directly against this PR's own CI run, not just by reading the YAML:

**The "Comment bundle size delta" step silently fails on this exact PR.** Job [85172914282](https://github.com/OpenHikmah/openhikmah-web/actions/runs/28721932187/job/85172914282), step "Comment bundle size delta", actual log output:
```
RequestError [HttpError]: Resource not accessible by integration
```
It shows green in the Checks tab only because `continue-on-error: true` masks the failure. I also checked this PR's comment list directly — no "Bundle Size Report" comment exists.

**Root cause:** GitHub always forces `GITHUB_TOKEN` to read-only for workflows triggered by `pull_request` from a **forked** repository, regardless of the `permissions:` block declared on the job (`issues: write` / `pull-requests: write` here). Since this PR is itself from a fork (`hfl0506/openhikmah-web`), and that's the normal case for outside contributors on an OSS project, this feature will silently no-op for essentially every external contribution — exactly the audience it matters most for.

**Suggested fix:** this needs a `pull_request_target`-based approach (or a two-workflow `workflow_run` pattern — build/measure on `pull_request`, upload the size numbers as an artifact, then a second privileged workflow triggered by `workflow_run` downloads the artifact and posts the comment). Either way, treat any PR-comment-posting step here as untested until it's been verified to actually post on a real fork PR — a green check on `continue-on-error: true` isn't evidence it works.

Everything else (the base-branch comparison logic, the `.size-limit.cjs` config, `persist-credentials: false` hardening) looks correct. Requesting changes just for the silently-broken comment step.

nazarli-shabnam:
Pushed a fix commit (`bd6651d`) directly to this branch (maintainer edits are enabled) to get this merge-ready:

**1. Merge conflict** — `main` picked up the e2e/a11y CI work (#137) while this PR was open. Resolved: both the new `e2e` job and this PR's bundle-size steps are kept in `ci.yml`'s `build` job; `package.json`/`bun.lock` keep both sets of new devDependencies (regenerated the lockfile via `bun install` rather than hand-editing it).

**2. The PR-comment step was still silently failing** — job [85172914282](https://github.com/OpenHikmah/openhikmah-web/actions/runs/28721932187/job/85172914282), step "Comment bundle size delta", was hitting `RequestError [HttpError]: Resource not accessible by integration` (403), masked by `continue-on-error: true`. This wasn't a config mistake — GitHub forces a **read-only** `GITHUB_TOKEN` on any `pull_request`-triggered workflow run that originates from a fork, no matter what `permissions:` the workflow declares. It's a hard platform restriction (this repo is fine to leave with a strict default `pull_request` trigger — `pull_request_target` would avoid the token restriction but runs untrusted fork code with write credentials, which is worse).

Fixed with GitHub's documented pattern for this: the `build` job now just uploads the size reports as an artifact (`bundle-size-reports`) plus the PR number; a new workflow, `bundle-size-comment.yml`, triggered on `workflow_run` (which runs in the base repo's context, so it gets a normal write token even for fork PRs), downloads that artifact and does the actual commenting.

One caveat worth knowing: GitHub only loads `workflow_run`-triggered workflows from the **default branch**, so `bundle-size-comment.yml` won't actually fire on this PR until it's merged — I confirmed via the Actions API that it isn't registered yet. `bun run size` (the actual budget gate — the important part of #136) already works correctly right now; the comment will start appearing on the *next* PR after this one merges.

Verified locally: `bun run build`, `bun run size` (498.3 kB vs. the 625 kB budget), `bun run lint`, `bun run typecheck` all pass. CI is green on the pushed commit. Thanks for the work on this — nice catch on the JSON-guard fix too.

nazarli-shabnam (approved):

PR #158, opened 2026-07-05
fix: validate shared-canvas node shape and harden the OG image handler

### LINKED ISSUE #118: fix: POST /api/share doesn't validate node shape, crashing OG image generation
## Describe the bug

`POST /api/share` (`app/api/share/route.ts`) validates only the outer canvas shape, not the shape of each node inside it:

```ts
// app/api/share/route.ts, lines 43-51
if (
  typeof body !== "object" ||
  body === null ||
  (body as { v?: unknown }).v !== 1 ||
  !Array.isArray((body as { nodes?: unknown }).nodes) ||
  (body as { nodes: unknown[] }).nodes.length === 0
) {
  return NextResponse.json({ error: "Invalid canvas" }, { status: 400 });
}
```

Nothing here checks that each entry in `nodes` has a `.verse` object, so `POST { "v": 1, "nodes": [{}] }` passes validation, gets stored, and returns a valid share `id`.

That id's Open Graph image is generated by `app/api/share/[id]/opengraph-image.tsx`, which then crashes on exactly this shape:

```ts
// app/api/share/[id]/opengraph-image.tsx, lines 33-53
const rows = await db
  .select()
  .from(sharedCanvases)
  .where(eq(sharedCanvases.id, id))
  .limit(1)
  .catch(() => []);

if (!rows[0]) return fallback();

const canvas = JSON.parse(rows[0].data) as SavedCanvas;   // <-- no try/catch
if (!canvas?.nodes?.length) return fallback();

const first = canvas.nodes[0];
const count = canvas.nodes.length;

return new ImageResponse(
  renderOgCard({
    eyebrow: `${count} verse${count === 1 ? "" : "s"}`,
    refPill: first.verse.ref,        // <-- `first.verse` is undefined here
    title: first.verse.surahName,
    body: clampBody(first.verse.translation),
  }),
  { ...OG_SIZE }
);
```

Two things stand out:
1. `JSON.parse(rows[0].data)` has no `try/catch` here, even though the sibling `GET` route for the same data (`app/api/share/[id]/route.ts:27-32`) wraps the identical parse in a `try/catch` specifically to handle corrupted data gracefully (`{ error: "Corrupted canvas data" }, { status: 500 }`).
2. `first.verse.ref` (and `.surahName`, `.translation`) are accessed with no existence check on `.verse`, even though this file already has a working `fallback()` function that's used for every *other* edge case in the same handler (missing row, empty `nodes` array, invalid UUID format).

So a canvas with a node lacking `.verse` — which the `POST` handler currently allows — throws an uncaught `TypeError` in the OG image handler instead of falling back to the generic card the code is clearly designed to produce for exactly this kind of malformed-data situation.

**Secondary, lower-severity issue in the same file/root cause**: the IP-based rate limiter guarding this endpoint (`app/api/share/route.ts:12`, `const ipBuckets = new Map<string, {...}>()`) is a module-level in-memory `Map`, unlike every other rate-limited route in the app (e.g. `app/api/workspace`, `app/api/social/activity`), which goes through the shared, Redis-backed `lib/rate-limit.ts`. On a multi-instance deployment this would enforce the stated "10/hour" limit independently per process, making the effective limit `10 × instance_count`. Low real-world impact today since the app deliberately runs single-box (per the project's own scaling notes), but worth fixing before that assumption changes, and it's the same "share endpoint isn't as hardened as the rest of the app" theme as the crash above.

**Tertiary note**: `app/api/share/route.ts:53-56`'s best-effort TTL-cleanup delete (`db.delete(sharedCanvases).where(...).catch(() => {})`) silently swallows any DB error with no log line, unlike every other DB error path elsewhere in the codebase, which logs via `console.error`. Not user-facing, but a real DB problem here (e.g. a permissions/schema issue) would be invisible in production logs.

## Steps to reproduce

1. `POST /api/share` with body `{"v":1,"nodes":[{}]}`.
2. Note the returned `id` in the `200` response.
3. Request `GET /api/share/<id>/opengraph-image`.

**Expected behaviour:** Either (a) step 1 rejects the malformed node with a `400`, since it doesn't have the shape any canvas node is supposed to have, or (b) if malformed data somehow does get stored, step 3 falls back to the generic OG card the same way it already does for every other malformed-data case in the same file.

**Actual behaviour:** Step 1 succeeds and stores the malformed canvas. Step 3 throws an uncaught `TypeError` reading `first.verse.ref` on `undefined`, instead of returning the fallback card.

## Environment

- Browser / OS: N/A — reproducible via direct HTTP requests; the user-facing symptom would be a broken link-preview image when a malformed shared-canvas link is posted anywhere that generates a preview (Slack, Twitter/X, iMessage, etc.)
- Node.js version: N/A
- Branch / commit: `main` @ current `HEAD`
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked — verified by direct code inspection of both handlers; not sent against the live deployment

Expected error shape: `TypeError: Cannot read properties of undefined (reading 'ref')` thrown inside the `opengraph-image.tsx` default export, uncaught.

## Additional context

- Files: `app/api/share/route.ts` (validation gap, lines 43-51; rate limiter, line 12; silent delete failure, lines 53-56), `app/api/share/[id]/opengraph-image.tsx` (unguarded parse + access, lines 42, 51-53).
- Suggested fixes:
  1. In `POST /api/share`, validate each node has a `verse` object with the fields the OG image (and presumably the canvas renderer itself) depends on, or at minimum reject nodes missing `.verse` entirely.
  2. In `opengraph-image.tsx`, wrap the `JSON.parse` in a `try/catch` (matching the sibling `GET` route) and guard `first.verse` before accessing its fields, falling back to `fallback()` on either failure — this is the more defensive, lower-risk fix and is worth doing even if (1) is also fixed, since external data (a manually crafted request bypassing normal canvas-save flow) shouldn't be able to crash a public-facing image handler.
  3. For the rate limiter, route through the existing shared `lib/rate-limit.ts` infrastructure instead of a module-level `Map`, matching every other rate-limited endpoint.
  4. For the silent delete failure, add a `console.error` in the `.catch(() => {})` at line 56, matching the logging convention used everywhere else in the codebase.

---

nazarli-shabnam:
## Summary

Fixes #118. `POST /api/share` only validated the outer `{v, nodes}` shape, so a node missing `.verse` (e.g. `{"v":1,"nodes":[{}]}`) was accepted and stored. The OG image handler (`app/api/share/[id]/opengraph-image.tsx`) then crashed reading `first.verse.ref` on `undefined`, instead of falling back like every other malformed-data case already handled in that file.

## Changes

- **`app/api/share/route.ts`**
  - Reject nodes whose `.verse` is missing, or missing the fields actually read downstream (`ref`, `surahName`, `translation`) — same 400/`"Invalid canvas"` response already used for the outer-shape check.
  - Replaced the module-level `ipBuckets` Map with the shared Redis-backed rate limiter (`lib/rate-limit.ts`, via `rateLimitOrNull`), matching every other rate-limited route (e.g. `app/api/workspace`). Same 10/hour budget as before — this was previously enforced independently per process, which under-enforces on a multi-instance deployment.
  - Logs the previously-silent TTL-cleanup delete failure (`.catch(() => {})` → `.catch((err) => console.error(...))`), matching the logging convention elsewhere in the codebase.
- **`app/api/share/[id]/opengraph-image.tsx`**
  - Wrapped `JSON.parse(rows[0].data)` in a `try/catch` → `fallback()`, matching the sibling `GET /api/share/[id]/route.ts`'s handling of the identical parse.
  - Guards `first?.verse` before reading its fields → `fallback()` if absent.
- **`__tests__/api/share.test.ts`** (new) — covers: 400 on a node missing/malformed `.verse`, 200 happy path (regression guard), 429 via a mocked rate limiter; and for the OG image handler, fallback (not a thrown error) on invalid JSON and on a node missing `.verse`, plus a normal-render regression guard.

## Testing
- `bun run test:ci` — full suite green (517/517), including the 7 new tests.
- `bun run typecheck` — clean.
- `bun run lint` / `bun run format:check` — clean on all touched files.
- Manually traced the issue's exact repro (`{"v":1,"nodes":[{}]}`) against the final code: `POST` now returns 400 instead of storing it; if such a row already existed from before this fix, the OG route now returns the fallback card instead of throwing.

### DIFF
--- a/__tests__/api/share.test.ts
+++ b/__tests__/api/share.test.ts
@@ -125,6 +125,15 @@ describe("GET /api/share/[id]/opengraph-image", () => {
     await expect(imageReq(VALID_ID)).resolves.toBeDefined();
   });
 
+  it("falls back instead of throwing when .verse is present but missing required fields", async () => {
+    mockSelect.mockReturnValue(
+      makeDbChain([
+        { id: VALID_ID, data: JSON.stringify({ v: 1, nodes: [{ verse: { ref: "2:255" } }] }) },
+      ])
+    );
+    await expect(imageReq(VALID_ID)).resolves.toBeDefined();
+  });
+
   it("renders normally for well-formed stored data", async () => {
     mockSelect.mockReturnValue(
       makeDbChain([
--- a/__tests__/api/share.test.ts
+++ b/__tests__/api/share.test.ts
@@ -1,0 +1,136 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest, NextResponse } from "next/server";
+
+// ── Mocks ──────────────────────────────────────────────────────────────────────
+
+function makeDbChain(resolveWith: unknown = []) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const chain: any = new Proxy(
+    function () {
+      return chain;
+    },
+    {
+      get(_t, prop) {
+        if (prop === "then")
+          return (res: (v: unknown) => unknown, rej?: (e: unknown) => unknown) =>
+            Promise.resolve(resolveWith).then(res, rej);
+        if (prop === "catch")
+          return (rej: (e: unknown) => unknown) => Promise.resolve(resolveWith).catch(rej);
+        if (prop === Symbol.toStringTag) return "MockChain";
+        return () => chain;
+      },
+      apply() {
+        return chain;
+      },
+    }
+  );
+  return chain;
+}
+
+const { mockSelect, mockInsert, mockDelete, mockRateLimitOrNull } = vi.hoisted(() => ({
+  mockSelect: vi.fn(() => makeDbChain([])),
+  mockInsert: vi.fn(() => makeDbChain([])),
+  mockDelete: vi.fn(() => makeDbChain([])),
+  mockRateLimitOrNull: vi.fn(async (): Promise<NextResponse | null> => null),
+}));
+
+vi.mock("@/lib/db", () => ({
+  db: {
+    select: mockSelect,
+    insert: mockInsert,
+    delete: mockDelete,
+  },
+}));
+vi.mock("@/lib/rate-limit", () => ({
+  rateLimitOrNull: mockRateLimitOrNull,
+}));
+
+import { POST } from "@/app/api/share/route";
+import Image from "@/app/api/share/[id]/opengraph-image";
+
+// ── Helpers ────────────────────────────────────────────────────────────────────
+
+const validVerse = {
+  surah: 2,
+  ayah: 255,
+  ref: "2:255",
+  arabicText: "arabic",
+  translation: "Allah - there is no deity except Him.",
+  surahName: "Al-Baqarah",
+  surahNameArabic: "البقرة",
+};
+
+function postReq(body: unknown) {
+  return new NextRequest("http://localhost/api/share", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(body),
+  });
+}
+
+const VALID_ID = "11111111-1111-1111-1111-111111111111";
+
+beforeEach(() => {
+  mockSelect.mockReset().mockReturnValue(makeDbChain([]));
+  mockInsert.mockReset().mockReturnValue(makeDbChain([]));
+  mockDelete.mockReset().mockReturnValue(makeDbChain([]));
+  mockRateLimitOrNull.mockReset().mockResolvedValue(null);
+});
+
+describe("POST /api/share", () => {
+  it("returns 400 when a node is missing .verse", async () => {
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
@coderabbitai review

nazarli-shabnam:
CodeRabbit's finding was valid — verified: \`first?.verse\` truthy-check alone didn't catch a stored node with a malformed \`verse\` (e.g. missing \`translation\`), which would still crash on \`clampBody(undefined).trim()\`. Fixed in 4201b21 by extracting the shape check already used by \`POST /api/share\`'s validation into \`lib/share-canvas.ts\` and reusing it here, so both entry points agree on what a valid node looks like. Added a regression test for exactly this case. Marked the thread resolved.

PR #163, opened 2026-07-07
fix(home): add heading to auth loading state

### LINKED ISSUE #150: fix: home page missing an <h1> heading (axe page-has-heading-one)
## Problem
The home page (`app/page.tsx`) has no `<h1>` element, which fails the axe-core `page-has-heading-one` rule. This was already surfaced during the e2e/a11y CI work (#137): `.superpowers/sdd/task-6-report.md` recorded it as a moderate violation when the a11y scan was first run against home/canvas/social.

It currently does **not** fail CI: `e2e/a11y.spec.ts` only hard-fails on `serious`/`critical` axe violations and treats `moderate`/`minor` findings (like this one) as advisory/warn-only. So this regression is invisible to the pipeline even though it's a confirmed, real accessibility gap.

## Proposed fix
- Add a semantic `<h1>` to the home page's render tree (for both the `PersonalHome` and `MarketingHero` variants, since the adaptive home renders one or the other — see `docs/plan.md` §2 "Adaptive home (PersonalHome / MarketingHero)").
- It doesn't need to be the dominant visual heading if that clashes with the existing hero design — it can be visually styled to match the current hero text while being the actual `<h1>` in the DOM, or use a visually-hidden (`sr-only`) `<h1>` if there's a strong design reason not to introduce a second visible heading-sized element.
- Re-run `bun run test:e2e` (specifically `e2e/a11y.spec.ts`) afterward and confirm the `page-has-heading-one` finding is gone from the home page scan.

## Why this matters
Screen reader users rely on a page's `<h1>` to identify its primary topic/landmark. This is a real WCAG/accessibility gap on the most-visited page in the app, not a false positive — it was independently confirmed via a live axe-core scan, not just linting.

ahfoysal:
## Summary
- add an `h1` to the authenticated loading skeleton on the home page
- cover the loading state heading in the component test

## Tests
- bun install --frozen-lockfile
- bun run format:check
- bun run lint
- bun run typecheck
- bun run test:ci
- bun run build
- targeted home a11y Playwright test
- pre-push integration tests

Fixes #150

### DIFF
--- a/__tests__/components/home/AuthLoadingSkeleton.test.tsx
+++ b/__tests__/components/home/AuthLoadingSkeleton.test.tsx
@@ -1,0 +1,11 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { AuthLoadingSkeleton } from "@/components/home/AuthLoadingSkeleton";
+
+describe("AuthLoadingSkeleton", () => {
+  it("keeps the home page heading available while auth restores", () => {
+    render(<AuthLoadingSkeleton />);
+
+    expect(screen.getByRole("heading", { level: 1, name: "Open Hikmah home" })).toBeInTheDocument();
+  });
+});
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Thanks for the contribution!

nazarli-shabnam (approved):

PR #164, opened 2026-07-08
test: add unit tests for lib/tafsir.ts sanitizer (#119)

### LINKED ISSUE #119: fix: lib/tafsir.ts (XSS sanitizer for external HTML) has zero test coverage
## Describe the bug

`lib/tafsir.ts`'s `htmlToTafsirBlocks`/`stripTags` functions are the **sole sanitization layer** between raw HTML fetched from an external API (quran.com's Ibn Kathir tafsir endpoint, via `app/api/verse/[surah]/[ayah]/tafsir/route.ts`) and what gets rendered in the UI (`components/layout/ContextSidebar.tsx`). The module's own doc comment makes this explicit:

```ts
// lib/tafsir.ts, lines 1-6
/**
 * Turns quran.com tafsir HTML into ordered plain-text blocks so the UI can render
 * it as escaped React text (no dangerouslySetInnerHTML, no XSS surface) while
 * keeping the structure: English commentary paragraphs and the quoted Arabic
 * phrases (`<div class="arabic …">`) it explains.
 */
```

`stripTags` is specifically written to defend against HTML-stripping bypass tricks — it decodes entities first (so an entity-encoded bracket can't survive as markup), then loops the tag-strip regex until the string stops changing (to catch reconstructed tags like `<scr<script>ipt>`), then does a final sweep removing any stray `<`/`>` characters. This is careful, security-relevant logic.

Despite that, there is **no test file for this module** — `__tests__/**/*tafsir*` (and any other reasonable naming pattern) returns no matches in the `__tests__/` directory, which otherwise has 44+ test files covering the rest of the codebase, including for far less security-sensitive logic. There's no regression protection for:
- The core guarantee itself (output never contains `<` or `>`, per the code's own reasoning).
- The entity-decode-then-strip ordering (getting this backwards would reopen the entity-encoded-bracket bypass the code is explicitly designed to prevent).
- The iterative strip-until-stable loop (a naive single-pass implementation would be vulnerable to the reconstructed-tag trick the loop exists to defeat).
- The `arabic: boolean` classification logic (`/class=["'][^"']*\barabic\b/i.test(attrs)`) that determines RTL/Amiri rendering.

I read the implementation directly and it looks correct today — this issue isn't reporting a bug in the sanitizer's current behavior, it's reporting that **the thing standing between external HTML and the rendered page has zero regression coverage**, so a future refactor (e.g. changing the strip-loop logic, or "simplifying" the entity-decode step) could silently reopen an XSS vector with no test to catch it.

## Steps to reproduce

N/A — this is a missing-test-coverage report, not a behavioral bug. To confirm the gap:
1. `git grep -l tafsir __tests__/` (or equivalent) — no results.
2. Compare against `lib/tafsir.ts`'s own doc comment explicitly claiming "no XSS surface" — a claim with no test asserting it.

**Expected behaviour:** A module explicitly responsible for preventing XSS from external content should have direct unit tests asserting the sanitization guarantee, the same way other security-relevant logic in this codebase is tested (e.g. `__tests__/lib/rate-limit.test.ts` exists for the rate limiter).

**Actual behaviour:** No test file exists for `lib/tafsir.ts`.

## Environment

- Browser / OS: N/A — test-coverage gap, not a runtime bug
- Node.js version: N/A
- Branch / commit: `main` @ current `HEAD`
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not applicable — this is a missing-tests report

N/A.

## Additional context

- File to add tests for: `lib/tafsir.ts`.
- Suggested test cases for `__tests__/lib/tafsir.test.ts`, following the existing Vitest conventions used elsewhere in `__tests__/`:
  - Plain paragraph/div/heading blocks are extracted with their text, tags stripped.
  - `class="...arabic..."` blocks are marked `arabic: true`; others `arabic: false`.
  - Entity-encoded content (`&lt;script&gt;`) decodes to literal text, not markup, and never survives as an unescaped `<`/`>` in the output.
  - A crafted reconstructed-tag string (e.g. `<scr<script>ipt>alert(1)</scr</script>ipt>`) is fully neutralized — no `<`/`>` in the output for any adversarial input, not just the happy path.
  - The fallback path (no recognized block tags) returns the whole input as one English block, still stripped of markup.
  - Empty/whitespace-only blocks are skipped (not pushed as empty-string entries).

---

k-anushka14:
## Summary
- Added `__tests__/lib/tafsir.test.ts` with unit tests covering `lib/tafsir.ts`'s HTML sanitization logic
- Tests verify the core XSS-prevention guarantee: no `<`/`>` survives in output, including against an entity-encoding bypass and a reconstructed-tag bypass attempt
- Closes #119

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [x] Tests

## AI / Theological changes

**Details** (if applicable):
N/A

## Testing
- [ ] `bun run test:ci` passes locally
- [ ] Tested in browser (for UI changes)

## Checklist
- [x] Conventional commits

### DIFF
--- a/__tests__/lib/tafsir.test.ts
+++ b/__tests__/lib/tafsir.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect } from "vitest";
+import { htmlToTafsirBlocks } from "@/lib/tafsir";
+
+describe("htmlToTafsirBlocks", () => {
+  it("extracts plain paragraph, div, and heading blocks with tags stripped", () => {
+    const blocks = htmlToTafsirBlocks("<h2>Title</h2><p>Body <b>text</b></p><div>Plain div</div>");
+    expect(blocks).toEqual([
+      { arabic: false, text: "Title" },
+      { arabic: false, text: "Body text" },
+      { arabic: false, text: "Plain div" },
+    ]);
+  });
+
+  it('marks class="...arabic..." blocks as arabic: true and others as false', () => {
+    const blocks = htmlToTafsirBlocks(
+      '<div class="foo arabic bar">quoted</div><div class="arabicish">not a match</div>'
+    );
+    // Word-boundary regex: "arabic" must appear as a whole class token, so
+    // "arabicish" must NOT be classified as arabic.
+    expect(blocks).toEqual([
+      { arabic: true, text: "quoted" },
+      { arabic: false, text: "not a match" },
+    ]);
+  });
+
+  it("decodes entity-encoded markup to literal text, never letting it survive as unescaped angle brackets", () => {
+    const blocks = htmlToTafsirBlocks("<div>&lt;script&gt;alert(1)&lt;/script&gt;</div>");
+    expect(blocks).toHaveLength(1);
+    expect(blocks[0].text).toBe("alert(1)");
+    expect(blocks[0].text).not.toMatch(/[<>]/);
+  });
+
+  it("neutralizes a reconstructed-tag bypass attempt (no < or > survives, for any adversarial input)", () => {
+    // The classic hide-a-tag-inside-a-tag trick: a single-pass strip would
+    // leave "<script>" behind once the outer layer is removed. The
+    // strip-until-stable loop must catch it.
+    const blocks = htmlToTafsirBlocks("<div><scr<script>ipt>alert(1)</scr</script>ipt></div>");
+    expect(blocks).toHaveLength(1);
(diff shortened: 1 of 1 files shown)


nazarli-shabnam (approved):
Thanks for your contribution!

PR #183, opened 2026-07-10
fix(names): log and count upstream Quran API fetch failures

### LINKED ISSUE #99: fix: Quran API fetch failures in names/verses route are silently swallowed
## Describe the bug

In `app/api/names/[slug]/verses/route.ts`, the two functions that fetch verse data from external Quran APIs — `fetchVerseData` (calls `api.alquran.cloud`) and `searchVerseRefs` (calls `api.quran.com`) — each wrap their network call in a `try/catch` that discards the error completely:

```ts
// fetchVerseData, lines 29-48
try {
  const [arabicRes, translationRes] = await Promise.all([
    fetch(`https://api.alquran.cloud/v1/ayah/${surahNum}:${ayahNum}/ar.alafasy`),
    fetch(`https://api.alquran.cloud/v1/ayah/${surahNum}:${ayahNum}/en.sahih`),
  ]);
  if (!arabicRes.ok || !translationRes.ok) return null;
  ...
} catch {
  return null;
}
```

```ts
// searchVerseRefs, lines 52-66
try {
  const url = `https://api.quran.com/api/v4/search?q=${encodeURIComponent(arabic)}&size=8&language=en&page=1`;
  const res = await fetch(url, { headers: { Accept: "application/json" }, next: { revalidate: 86400 } });
  if (!res.ok) return [];
  ...
} catch {
  return [];
}
```

Neither catch block logs anything or increments a metrics counter. This is inconsistent with the rest of the same file: `buildReasons` (line 93-97) and `fallbackAIVerses` (line 117-121) both `console.error` on failure specifically so "a persistently malformed AI response is visible, not silent" (per their own inline comments). The two functions that talk to the *external* Quran APIs — arguably the more failure-prone dependency, since they're outbound HTTP calls to third-party services rather than local AI SDK calls — have no equivalent safety net.

Downstream, `getVersesBySlug` (line 129) treats a `null`/`[]` result from these functions exactly the same as "this name legitimately has no verses to show" — it falls through to the AI-only fallback, and if that also comes up empty, returns `[]`. The route handler returns `200 []`, a fully successful HTTP response, so nothing downstream can distinguish "no data" from "upstream is down."

## Steps to reproduce

1. Temporarily block or fail requests to `api.alquran.cloud` and `api.quran.com` (e.g. via a local proxy, or by observing behavior during a real outage of either service).
2. Request `GET /api/names/<any-slug>/verses` for a name whose content isn't already cached (or bump `VERSES_VERSION` to force regeneration).
3. Observe the response and server logs.

**Expected behaviour:** A `200 []` response is only possible for a genuine "this name has no associated verses" case. When the failure is caused by an upstream API being unreachable, rate-limited, or erroring, the server logs should show a clear error message, and ideally a metrics counter (`/api/metrics`) should reflect the failure so it can be distinguished from steady-state.

**Actual behaviour:** The response is `200 []` either way, and nothing is logged — an upstream outage is indistinguishable from "no verses found" both to the client and to anyone watching server logs or `/api/metrics`.

## Environment

- Browser / OS: N/A — server-side route handler bug, reproducible via direct API request from any environment
- Node.js version: N/A
- Branch / commit: `main` @ `ba8e63b` (current `HEAD` at time of filing)
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (requires forcing an upstream outage to observe; the code path is confirmed by reading, not by live reproduction)

N/A — the bug is the *absence* of console output, not an error message.

## Additional context

- File: `app/api/names/[slug]/verses/route.ts`
  - `fetchVerseData`: lines 46-48
  - `searchVerseRefs`: lines 63-65
- There's also a secondary effect worth noting: because `getOrGenerateNameContent` only skips caching when the result `isEmpty` (line 177: `(v) => v.length === 0`), an empty result from a transient outage is *not* cached — which is good (it'll retry next request) but also means a sustained outage causes the app to keep re-hitting the failing upstream API on every request for that name, with no visibility into how often that's happening.
- Suggested fix, matching the existing pattern already used elsewhere in this same file (see `buildReasons`/`fallbackAIVerses`) and the `incr()` metrics pattern used in `lib/rate-limit.ts` (`ratelimit_fail_open`, `ratelimit_redis_fallback`):
  ```ts
  } catch (err) {
    console.error(`Name verses: alquran.cloud fetch failed for ${ref}:`, err);
    incr("quran_api_fetch_error");
    return null;
  }
  ```
  (and equivalently for `searchVerseRefs`, e.g. `incr("quran_search_api_error")`), so an upstream outage becomes visible on `/api/metrics` the same way other degraded-mode paths already are in this codebase.

---

nazarli-shabnam:
## Summary
- `fetchVerseData` (api.alquran.cloud) and `searchVerseRefs` (api.quran.com) in `app/api/names/[slug]/verses/route.ts` swallowed fetch errors completely — no log, no metric — so an upstream outage was indistinguishable from a genuine "no verses found" result (both returned `200 []`).
- Both catch blocks now `console.error` the failure and increment a metrics counter (`quran_api_fetch_error`, `quran_search_api_error`), matching the existing pattern used by `buildReasons`/`fallbackAIVerses` in the same file and the `incr()` convention used in `lib/infra/rate-limit.ts`.

## Test plan
- [x] `npm run typecheck` passes
- [x] `eslint` on the changed file passes with no warnings
- [x] Pre-commit hook (lint, typecheck, unit tests) passed
- [x] Pre-push hook (integration tests) passed
- [ ] Manual: force an upstream failure (e.g. block `api.alquran.cloud`/`api.quran.com`) and confirm `console.error` output and `quran_api_fetch_error`/`quran_search_api_error` appear on `/api/metrics`

Fixes #99

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved monitoring for failures when retrieving Quran verse data or search results.
  * Errors are now recorded in logs and metrics while preserving existing fallback behavior.

### DIFF
--- a/app/api/names/[slug]/verses/route.ts
+++ b/app/api/names/[slug]/verses/route.ts
@@ -3,6 +3,7 @@ import { callAI } from "@/lib/ai/ai";
 import { getNameBySlug } from "@/lib/names/divine-names";
 import { getSurahName } from "@/lib/quran/surah-names";
 import { getOrGenerateNameContent } from "@/lib/names/name-content";
+import { incr } from "@/lib/infra/metrics";
 import sanitizeHtml from "sanitize-html";
 import type { VerseRef } from "@/types/quran";
 
@@ -46,7 +47,11 @@ async function fetchVerseData(ref: string): Promise<Omit<NameVerse, "reason"> | 
       surahName,
       surahNameArabic,
     };
-  } catch {
+  } catch (err) {
+    // An upstream outage must not look identical to "no verse data" —
+    // log and count it so it's visible on /api/metrics, not silent.
+    console.error(`Name verses: alquran.cloud fetch failed for ${ref}:`, err);
+    incr("quran_api_fetch_error");
     return null;
   }
 }
@@ -66,7 +71,11 @@ async function searchVerseRefs(arabic: string): Promise<string[]> {
       .filter((r) => r.verse_key && /^\d+:\d+$/.test(r.verse_key))
       .map((r) => r.verse_key as string)
       .slice(0, 5);
-  } catch {
+  } catch (err) {
+    // Same reasoning as fetchVerseData: a failed search must be distinguishable
+    // from "genuinely no matching verses" in logs and /api/metrics.
+    console.error(`Name verses: quran.com search failed for "${arabic}":`, err);
+    incr("quran_search_api_error");
     return [];
   }
 }


nazarli-shabnam:
@coderabbitai review

Issue #126, opened 2026-07-03
Apply rate limiting to AI-expansion endpoints


nazarli-shabnam:
## Problem
`SECURITY.md` explicitly documents AI-expansion-call rate limiting as an accepted/out-of-scope gap. Meanwhile, `lib/rate-limit.ts` (backed by `lib/redis.ts` and a `rateLimits` DB table) is already applied across other routes — workspace, `social/*`, notes, search, share, connections — but not on the AI-expansion API routes themselves.

## Proposed fix
- Apply the existing `lib/rate-limit.ts` utility to the AI-expansion route(s).
- Choose limits appropriate to the cost of the underlying AI calls (likely stricter than the general-purpose routes, given the cost-per-call and DoS/cost-abuse risk).
- Update `SECURITY.md` to remove/update the "accepted gap" note once closed.

## Why this matters
AI-expansion calls are typically far more expensive (latency + $) per request than typical CRUD routes, making them the most attractive target for abuse/cost-DoS if left unlimited.

PR #228, opened 2026-07-14
fix: cap arabicText/translation length in /api/connections

### LINKED ISSUE #172: fix: /api/connections forwards unbounded text into AI generation
## Problem
`app/api/connections/route.ts` POST forwards `arabicText`/`translation` body fields directly into `getConnections` (AI generation) with no length cap or sanitization before the AI call.

## Proposed fix
Add a reasonable max-length validation on both fields before they're used to build the AI prompt, returning 400 on violation. Consider basic sanitization/escaping if the fields are interpolated directly into a prompt template.

## Why this matters
Unbounded text forwarded into an AI call increases per-request cost (larger prompts = more tokens) and is a distinct DoS/cost-abuse and prompt-injection surface from the AI-expansion rate-limiting gap already tracked in #126 — this is about the payload itself, not request frequency.

nazarli-shabnam:
## Problem
\`app/api/connections/route.ts\` POST forwards \`arabicText\`/\`translation\` body fields directly into \`getConnections\` (AI generation) with no length cap or sanitization before the AI call — an unbounded-prompt cost-abuse and prompt-injection surface, distinct from the AI-expansion rate-limiting gap tracked separately in #126.

Closes #172

## Fix
Added a \`MAX_TEXT_LENGTH = 5000\` cap on both \`arabicText\` and \`translation\`, checked right after the existing \`fromRef\` validation and before the AI call. 5,000 chars is generous for even the longest verses (real verse text/translations are at most a few hundred characters) while closing the unbounded-payload gap. Mirrors the size-cap pattern this file already uses for \`excludeRefs\` (\`MAX_EXCLUDE_REFS\`).

## Testing
- Added two cases to \`__tests__/api/connections.test.ts\`: 400 when \`arabicText\` exceeds the cap (and confirms the rate-limit/AI path is never reached), 400 when \`translation\` exceeds the cap.
- \`bun run typecheck\`, \`bun run lint\`, and the full \`bun run test\` suite pass.

### DIFF
--- a/__tests__/api/connections.test.ts
+++ b/__tests__/api/connections.test.ts
@@ -154,6 +154,29 @@ describe("POST /api/connections", () => {
     expect(mockConsume).not.toHaveBeenCalled();
   });
 
+  it("returns 400 when arabicText exceeds the max length, without calling the AI path", async () => {
+    const req = makeRequest({
+      fromRef: "1:1",
+      kind: "thematic",
+      arabicText: "x".repeat(5001),
+      translation: "trans",
+    });
+    const res = await POST(req);
+    expect(res.status).toBe(400);
+    expect(mockConsume).not.toHaveBeenCalled();
+  });
+
+  it("returns 400 when translation exceeds the max length", async () => {
+    const req = makeRequest({
+      fromRef: "1:1",
+      kind: "thematic",
+      arabicText: "text",
+      translation: "x".repeat(5001),
+    });
+    const res = await POST(req);
+    expect(res.status).toBe(400);
+  });
+
   it("returns 400 for a malformed fromRef", async () => {
     const req = makeRequest({
       fromRef: "garbage",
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Reviewed this against #172.

`MAX_TEXT_LENGTH = 5000` is checked on both `arabicText` and `translation` right after `fromRef` validation and before `getConnections` (the AI call) and before rate-limit consumption — so an oversized payload is rejected before it costs anything, which is exactly the gap #172 described. The cap itself is generous relative to real verse/translation lengths (mirrors the existing `MAX_EXCLUDE_REFS` sizing pattern in the same file) while still closing the unbounded-prompt surface. Tests confirm the 400 response and that the AI/rate-limit path is never reached for the oversized-arabicText case. CI is fully green.

No gaps versus #172. Ready to merge — #228.

PR #229, opened 2026-07-14
fix: reject JWTs with missing or non-numeric exp claim

### LINKED ISSUE #107: fix: JWT accepted when exp claim is missing or non-numeric
## Describe the bug

`verifiedJwtSub()` in `lib/social-auth.ts` only rejects a token as expired when the `exp` claim is present **and** numeric:

```ts
// lib/social-auth.ts, line 197
if (typeof payload.exp === "number" && payload.exp * 1000 <= Date.now()) return null;
```

If `payload.exp` is absent entirely, or present as a non-numeric value, this condition is simply `false` and the function proceeds as if the token has no expiry at all — subject to passing the remaining checks (signature verification, `sub` present). Once accepted, the resolved user is cached in the in-process L1 cache and the shared Redis L2 cache for `CACHE_TTL_SECONDS` (5 minutes, per `lib/social-auth.ts:58`), so a token in this state would keep authenticating for at least that cache window regardless of what happens to it upstream.

This stands out against the rest of the function, which otherwise fails closed on every ambiguous case: unparseable header/payload → reject (line 188-190), unrecognized `alg` → reject (line 194), unverifiable signature → reject (line 223). The `exp` check is the one place that treats "can't determine" as "assume valid" instead of "assume invalid."

## Steps to reproduce

This requires a JWT signed by QF's real key material but lacking a numeric `exp` claim, which I can't manufacture from inside this repo (I don't hold QF's signing key) — this is a code-inspection finding based on reading the conditional directly, not an observed live token. Conceptually:

1. Obtain (or imagine) a QF-issued RS256-signed access token whose payload has no `exp` field, or a malformed one (e.g. a string instead of a number).
2. Present it as `Authorization: Bearer <token>` to any OpenHikmah API route that calls `requireUser`.
3. `verifiedJwtSub()` would evaluate `typeof payload.exp === "number"` as `false`, skip the expiry rejection, and proceed to signature verification — accepting the token indefinitely if the signature is valid.

**Expected behaviour:** A token without a valid, numeric `exp` claim should be treated as invalid (matching the fail-closed pattern used for every other check in this function), falling back to the authoritative userinfo path.

**Actual behaviour:** A missing or malformed `exp` claim silently skips expiry validation entirely.

**Confidence caveat:** under normal OAuth2/OIDC behavior, access tokens issued by a compliant authorization server always carry `exp`, so this is likely low real-world risk in practice — QF's Hydra-based token issuance would need to genuinely omit `exp` (or emit a non-numeric one) for this to be reachable, which I have no evidence it currently does. Flagging it because it's a one-line, zero-risk fix and because it's the only place in this function that doesn't already fail closed the way everything else does.

## Environment

- Browser / OS: N/A — server-side auth bug
- Node.js version: N/A
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked — requires a QF-issued token lacking `exp`, which I have no evidence QF currently produces; verified by reading the conditional's logic, not by triggering it live

N/A — no error occurs; this is a missing-check finding.

## Additional context

- File: `lib/social-auth.ts`, line 197.
- Suggested fix, matching the fail-closed convention used everywhere else in this function:
  ```ts
  if (typeof payload.exp !== "number" || payload.exp * 1000 <= Date.now()) return null;
  ```
- Related: this is the same function as the audience/issuer gap reported separately — worth fixing together since both are small, additive changes to the same block of code.

---

> Filed publicly at the repo owner's request (overriding the default private-disclosure routing in [SECURITY.md](../../SECURITY.md) for this specific, low-real-world-risk, confirmation-pending finding).

nazarli-shabnam:
## Problem
\`verifiedJwtSub()\` in \`lib/auth/social-auth.ts\` only rejected a token as expired when the \`exp\` claim was present **and** numeric:

\`\`\`ts
if (typeof payload.exp === "number" && payload.exp * 1000 <= Date.now()) return null;
\`\`\`

If \`exp\` is absent or non-numeric, this condition is simply \`false\` and the function proceeds as if the token has no expiry — the only place in this function that treats "can't determine" as "assume valid" instead of "assume invalid" (unparseable header/payload, unrecognized \`alg\`, and unverifiable signature all fail closed already).

Closes #107

## Fix
One-line fail-closed fix, exactly as the issue specifies:
\`\`\`ts
if (typeof payload.exp !== "number" || payload.exp * 1000 <= Date.now()) return null;
\`\`\`

## Testing
- Added two cases to \`__tests__/lib/auth/social-auth.test.ts\`: a validly-signed JWT with \`exp\` omitted entirely is rejected (401), and one with a non-numeric \`exp\` is rejected (401) — both previously would have proceeded to signature verification and succeeded.
- \`bun run typecheck\`, \`bun run lint\`, and the full \`bun run test\` suite pass.

### DIFF
--- a/__tests__/lib/auth/social-auth.test.ts
+++ b/__tests__/lib/auth/social-auth.test.ts
@@ -98,6 +98,20 @@ describe("requireUser — JWT signature verification", () => {
     expect("status" in res && (res as Response).status).toBe(401);
   });
 
+  it("rejects a validly-signed token whose payload omits exp entirely (fails closed)", async () => {
+    mockLimit.mockResolvedValue([user]); // even if a row existed, we must not reach it
+    const token = makeJwt({ sub: "qf-sub-123" });
+    const res = await requireUser(reqWith(token));
+    expect("status" in res && (res as Response).status).toBe(401);
+  });
+
+  it("rejects a validly-signed token whose exp is non-numeric (fails closed)", async () => {
+    mockLimit.mockResolvedValue([user]);
+    const token = makeJwt({ sub: "qf-sub-123", exp: "not-a-number" });
+    const res = await requireUser(reqWith(token));
+    expect("status" in res && (res as Response).status).toBe(401);
+  });
+
   it("rejects the alg=none bypass", async () => {
     mockLimit.mockResolvedValue([user]);
     const token = makeJwt({ sub: "qf-sub-123", exp: farFuture }, { alg: "none", sign: false });
--- a/lib/auth/social-auth.ts
+++ b/lib/auth/social-auth.ts
@@ -245,8 +245,10 @@ async function verifiedJwtSub(token: string): Promise<string | null> {
   // are the classic JWT signature-bypass vectors.
   if (header.alg !== "RS256") return null;
 
-  // Reject expired tokens.
-  if (typeof payload.exp === "number" && payload.exp * 1000 <= Date.now()) return null;
+  // Reject expired tokens. Fail closed like every other check in this
+  // function: a missing or non-numeric exp is treated as invalid, not as
+  // "no expiry."
+  if (typeof payload.exp !== "number" || payload.exp * 1000 <= Date.now()) return null;
 
   const keys = await fetchJwks();
   if (keys.length === 0) return null; // can't verify → caller uses userinfo


nazarli-shabnam:
Reviewed this against #107.

The one-line fix matches the issue's suggested fail-closed fix exactly: `typeof payload.exp !== "number" || payload.exp * 1000 <= Date.now()` now rejects a token whenever `exp` is missing or non-numeric, instead of silently treating it as "no expiry" — bringing this check in line with every other check in `verifiedJwtSub()`, which already failed closed (unparseable header/payload, unrecognized `alg`, unverifiable signature). Tests cover both cases the issue called out: `exp` omitted entirely, and `exp` as a non-numeric value — both correctly return 401 now. CI is fully green.

No gaps versus #107. Ready to merge — #229.

PR #232, opened 2026-07-14
fix: remove redundant decodeURIComponent on already-decoded Next.js params

### LINKED ISSUE #104: fix: unguarded decodeURIComponent throws uncaught 500 on malformed root/bookmark refs
## Describe the bug

Two route handlers call `decodeURIComponent` on an already-decoded Next.js dynamic route param, outside of any `try/catch`, so a malformed input crashes the handler with an unhandled `URIError` instead of returning the route's intended graceful response.

Next.js's App Router already URL-decodes dynamic segment values before they reach `params` (this is standard Next.js routing behavior — the raw path segment is decoded once by the framework). Both routes below decode it a *second* time, and in both cases the call sits before the `try` block that's meant to catch failures:

**`app/api/root/[root]/route.ts:19-26`:**
```ts
export async function GET(
  _req: NextRequest,
  { params }: { params: Promise<{ root: string }> }
) {
  const { root: rawRoot } = await params;
  const root = decodeURIComponent(rawRoot ?? "").trim();   // <-- outside try

  if (!root) {
    return NextResponse.json({ error: "Missing root" }, { status: 400 });
  }

  try {
    ...
```
This route's own doc comment (lines 7-11) states it "Returns an empty list (200) on any failure so the word popover degrades gracefully" — but that safety net is inside the `try` block below, and this decode happens before it, so this specific failure mode bypasses the documented graceful-degradation behavior entirely.

**`app/api/bookmarks/[ref]/route.ts:14-17`:**
```ts
export async function DELETE(
  req: NextRequest,
  { params }: { params: Promise<{ ref: string }> }
) {
  const authed = await requireUser(req);
  if (authed instanceof NextResponse) return authed;

  const { ref } = await params;
  const verseRef = decodeURIComponent(ref);   // <-- outside try

  try {
    await db.delete(bookmarks)...
```

In both cases, if the (already-decoded) segment contains a literal `%` not followed by two valid hex digits, `decodeURIComponent` throws `URIError: URI malformed`. Since the call is outside the `try`, the exception propagates out of the route handler entirely, which Next.js turns into an unhandled 500 error (with a server-side stack trace) instead of the handler's own designed error response.

Concretely: a client requests `GET /api/root/%25`. Next.js decodes the URL-encoded path segment `%25` → `%` before handing it to the route as `params.root`. The route then calls `decodeURIComponent("%")`, which throws (a bare `%` is not valid percent-encoding), uncaught.

For comparison, other dynamic-param routes in this codebase (e.g. `app/api/verse/[surah]/[ayah]/route.ts`, per the sibling routes referenced in `app/api/root/[root]/route.ts`'s own pattern) don't re-decode already-decoded Next.js params, since it's redundant — Next.js has already done it.

## Steps to reproduce

1. Send `GET /api/root/%25` (or any request where the raw path segment, once decoded once by Next.js, still contains a lone `%` not part of a valid `%XX` escape).
2. Observe the response.
3. Similarly, send `DELETE /api/bookmarks/%25` with a valid auth token.

**Expected behaviour:** `app/api/root/[root]/route.ts` should return its documented graceful degradation (`{ root, count: 0, verses: [] }` with status 200, matching its own `catch` block's behavior for other failures) or, at minimum, a clean 400. `app/api/bookmarks/[ref]/route.ts` should return a 400 for a malformed ref rather than crashing.

**Actual behaviour:** Both routes throw an uncaught `URIError` and return an unhandled 500 with a stack trace, bypassing their own intended error-handling paths.

## Environment

- Browser / OS: N/A — reproducible via any HTTP client
- Node.js version: N/A
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (verified by code inspection and `decodeURIComponent` semantics; not yet sent against the live deployment)

Expected server-side error shape: `URIError: URI malformed` thrown from `decodeURIComponent`, uncaught, surfaced by Next.js as a generic 500.

## Additional context

- Files: `app/api/root/[root]/route.ts:20`, `app/api/bookmarks/[ref]/route.ts:15`.
- Suggested fix — since Next.js already decodes the param, the simplest fix is to drop the redundant `decodeURIComponent` call entirely (matching the pattern already used by sibling dynamic-param routes that don't re-decode). If a second decode is genuinely needed for some reason specific to these routes, at minimum move it inside the existing `try` block so malformed input degrades the same way other failures already do in each of these handlers.

---

Mubashir78:
## Description

Removes redundant `decodeURIComponent` calls in two route handlers where Next.js has already URL-decoded the dynamic param value.

Next.js App Router decodes URL-encoded path segments before passing them to `params`. A second `decodeURIComponent` call throws `URIError: URI malformed` on bare `%` sequences (e.g. `GET /api/root/%25` where Next.js decodes `%25` → `%`, then the route crashes on `decodeURIComponent("%")`).

### Routes affected:

- `app/api/root/[root]/route.ts`: concordance endpoint that doc-comments promise "Returns an empty list (200) on any failure", but the decode sits before the `try/catch`, bypassing that graceful degradation.
- `app/api/bookmarks/[ref]/route.ts`: the decode sits before the `try` block, so a malformed ref crashes with an unhandled 500 instead of the intended error response.

Fixes #104

### DIFF
--- a/app/api/bookmarks/[ref]/route.ts
+++ b/app/api/bookmarks/[ref]/route.ts
@@ -9,7 +9,7 @@ export async function DELETE(req: NextRequest, { params }: { params: Promise<{ r
   if (authed instanceof NextResponse) return authed;
 
   const { ref } = await params;
-  const verseRef = decodeURIComponent(ref);
+  const verseRef = ref;
 
   try {
     await db
--- a/app/api/root/[root]/route.ts
+++ b/app/api/root/[root]/route.ts
@@ -14,7 +14,7 @@ const MAX_VERSES = 24;
 
 export async function GET(_req: NextRequest, { params }: { params: Promise<{ root: string }> }) {
   const { root: rawRoot } = await params;
-  const root = decodeURIComponent(rawRoot ?? "").trim();
+  const root = (rawRoot ?? "").trim();
 
   if (!root) {
     return NextResponse.json({ error: "Missing root" }, { status: 400 });


nazarli-shabnam:
Reviewed this against #104.

**Fix: correct and matches the issue's own recommended approach.** The redundant second `decodeURIComponent` call is dropped in both `app/api/root/[root]/route.ts` and `app/api/bookmarks/[ref]/route.ts`, since Next.js's App Router already decodes dynamic path segments once before `params` reaches the handler. I checked both routes end-to-end — the resulting (once-decoded) value is used consistently with how it's compared/stored downstream in each handler, so this doesn't introduce any encoding mismatch. This is exactly the "simplest fix" the issue itself suggested, and it resolves the `URIError: URI malformed` crash on inputs like `%25` for both routes.

So: the fix itself looks complete and correct

nazarli-shabnam (approved):

PR #233, opened 2026-07-14
fix: add metrics counter for admin audit log write failures

### LINKED ISSUE #101: fix: admin audit log writes can silently fail with no visibility
## Describe the bug

`logAdminAction()` in `lib/admin-audit.ts` is the single write path for the admin audit log — the comment on the function describes it as the mechanism by which "the console polices itself (who did what, when, to which target)." It is intentionally best-effort by design (per its own doc comment, "a logging failure must never break the action it records"), which is a reasonable choice — but the failure mode when it *does* fail is currently invisible:

```ts
// lib/admin-audit.ts, lines 26-44
export async function logAdminAction(entry: {
  adminQfId: string;
  action: string;
  targetType?: string | null;
  targetId?: string | null;
  meta?: unknown;
}): Promise<void> {
  try {
    await db.insert(adminAuditLog).values({ ... });
  } catch (err) {
    console.error("Failed to write admin audit log:", entry.action, err);
  }
}
```

On failure, the only trace is a single `console.error` line — no metrics counter, no alert, no retry, and (per the codebase's own pattern in `lib/rate-limit.ts`, which increments `ratelimit_fail_open` / `ratelimit_redis_fallback` counters whenever it degrades) no way to see on `/api/metrics` that audit coverage has a gap.

Compounding this, every admin route handler that calls `logAdminAction` does so **after** the privileged mutation has already committed, as a separate, non-transactional statement — e.g. `app/api/admin/users/route.ts:67-90` and `app/api/admin/connections/route.ts:67-83` both commit the mutation first, then call `logAdminAction` independently. If the audit `INSERT` fails for any transient reason (a connection drop immediately after the mutation's connection is returned to the pool, brief pool exhaustion, a constraint issue), the privileged action itself has already succeeded and the route returns `200` — but no audit row exists for it, and nothing surfaces that fact anywhere except a log line that may or may not be retained/monitored.

Given the entire purpose of this table is accountability ("who did what, when, to which target"), a silently-droppable write with no visibility is a meaningful gap in exactly the mechanism meant to catch that kind of gap.

## Steps to reproduce

1. Trigger any admin mutation that calls `logAdminAction` afterward (e.g. `PATCH /api/admin/users/[id]` to disable a user, or a connection retirement in `app/api/admin/connections`).
2. Introduce a transient DB failure specifically on the audit-log `INSERT` — e.g. a brief connection drop or pool exhaustion timed to hit after the primary mutation's transaction has already committed.
   - Difficult to force deterministically without fault-injecting the DB connection; the bug is confirmed by code inspection of the non-transactional write pattern and the empty-handed catch block, not by an observed production incident.
3. Confirm the admin mutation succeeded (e.g. the user is disabled) while checking `admin_audit_log` for the corresponding row.

**Expected behaviour:** Either (a) an operator-visible signal (metrics counter, alert) fires when an audit write fails, so the gap is known and can be investigated/backfilled, or (b) the audit write happens atomically with the action it records, so the two can never diverge.

**Actual behaviour:** A privileged action can succeed while its audit record silently fails to persist, with no signal beyond a console log line that isn't tracked anywhere.

## Environment

- Browser / OS: N/A — server-side bug
- Node.js version: N/A
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (requires fault-injecting a DB failure timed between the mutation commit and the audit insert; verified by code inspection)

N/A — the bug is that the only signal is an untracked `console.error` line, which is the problem being reported.

## Additional context

- File: `lib/admin-audit.ts`, catch block at lines 41-43.
- Example non-transactional call sites: `app/api/admin/users/route.ts:67-90`, `app/api/admin/connections/route.ts:67-83` (not exhaustive — this is the shape of every admin route that calls `logAdminAction`).
- Suggested minimal fix (matches the existing `incr()` pattern used elsewhere in the codebase for degraded-mode visibility, e.g. `ratelimit_fail_open` in `lib/rate-limit.ts`):
  ```ts
  } catch (err) {
    console.error("Failed to write admin audit log:", entry.action, err);
    incr("admin_audit_write_failed");
  }
  ```
- Stronger fix (bigger change, worth considering separately): write the audit row inside the same DB transaction as the mutation it records, so the action and its audit entry commit or roll back together and can never diverge. This would require restructuring each admin route handler to open an explicit transaction spanning both statements, so it's a larger change than the metrics-visibility fix above.

---

Mubashir78:
## Description

Adds an `incr("admin_audit_write_failed")` metric counter to the `catch` block in `logAdminAction` so that failures in admin audit log writes are visible on `/api/metrics`, matching the existing pattern used in `lib/infra/rate-limit.ts` (e.g. `ratelimit_fail_open`, `ratelimit_redis_fallback`).

With this change, operators can detect gaps via metrics even when no console monitoring is in place.

Fixes #101

### DIFF
--- a/lib/admin/admin-audit.ts
+++ b/lib/admin/admin-audit.ts
@@ -1,5 +1,6 @@
 import { db } from "@/lib/infra/db";
 import { adminAuditLog } from "@/lib/infra/db/schema";
+import { incr } from "@/lib/infra/metrics";
 
 /**
  * Serialize `meta` defensively: a BigInt or circular reference must not throw and
@@ -40,5 +41,6 @@ export async function logAdminAction(entry: {
     });
   } catch (err) {
     console.error("Failed to write admin audit log:", entry.action, err);
+    incr("admin_audit_write_failed");
   }
 }


nazarli-shabnam:
Thanks — verified this against #101.

`incr("admin_audit_write_failed")` is added to the existing catch block in `logAdminAction`, matching the exact minimal fix suggested in the issue and the codebase's existing `ratelimit_fail_open`/`ratelimit_redis_fallback` pattern in `lib/infra/rate-limit.ts`. (The issue's "stronger fix" — transactional audit writes — is explicitly called out as a separate, larger change and correctly left out of scope here.) CI is fully green.

No gaps — this fully resolves #101 as scoped. 👍

nazarli-shabnam (approved):

PR #234, opened 2026-07-14
fix: prevent audio player state desync on rapid track switches

### LINKED ISSUE #102: fix: audio player play/pause state desyncs after rapid track switches
## Describe the bug

`store/audio.ts` keeps a single module-level `HTMLAudioElement` (`_audio`, line 32) and drives all playback through `loadAndPlay()` (lines 41-47), which assigns a new `src`, calls `.load()`, and returns the `.play()` promise. Every entry point that starts a new track — `playVerse` (57-69), `playGraph` (71-85), `next` (114-126), `prev` (128-137) — follows the same shape:

```ts
set({ ..., isPlaying: true, isLoading: true, ... });
loadAndPlay(verse, () => get()._onEnded())
  .then(() => set({ isLoading: false }))
  .catch(() => set({ isPlaying: false, isLoading: false }));
```

There is no request-generation counter or sequence token guarding these `.then`/`.catch` callbacks. If a second play call happens before the first one's `play()` promise has settled — e.g. the user hits "Next" twice quickly, or taps a verse's play button while another verse is still loading — assigning the new `src` and calling `.load()` on the shared `_audio` element causes the browser to reject the *first* pending `play()` promise with an `AbortError` ("The play() request was interrupted by a new load request"). That stale rejection then runs its `.catch()`, which unconditionally sets `isPlaying: false` — even though the *second* track's `play()` call has since resolved and audio is actively playing.

`MiniPlayer` derives its play/pause icon purely from the `isPlaying` boolean (`components/audio/MiniPlayer.tsx:55,60`), so the UI now shows "Paused" while audio is audibly playing. Pressing the (visually "Play") button in this state calls `resume()` on an already-playing track, which is a no-op from the audio engine's perspective but confusing from the user's.

## Steps to reproduce

1. Start playing any verse (e.g. via a name's "verses" panel or canvas audio).
2. While it's loading/playing, click "Next" (or trigger another `playVerse`/`next`/`prev` call) twice in quick succession (within roughly 300ms — fast enough that the first `play()` promise hasn't settled before the second `loadAndPlay` reassigns `src`).
3. Observe the `MiniPlayer` icon and the actual audio state.

**Expected behaviour:** After two rapid track switches, the player UI reflects the actual state of the currently-loaded track (playing, since the second `play()` call succeeded).

**Actual behaviour:** The UI shows the paused/play icon (because the stale first call's `.catch()` set `isPlaying: false`), while audio for the second track is actually playing. The play/pause control is now desynced from reality until the user manually toggles it (which then calls `resume()` on already-playing audio, a confusing no-op).

## Environment

- Browser / OS: Any browser supporting the standard `HTMLMediaElement.play()` promise rejection behavior on interrupted load (Chrome, Firefox, Safari all implement this per spec) — not browser-specific.
- Node.js version: N/A — client-side store bug
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (verified by code inspection of the race condition; not yet reproduced against the live deployed site)

The browser console will typically show an unhandled/caught `AbortError` similar to: `Uncaught (in promise) DOMException: The play() request was interrupted by a new load request. https://goo.gl/LdLk22` — though since it's caught here, it surfaces only as the `isPlaying` state flip, not a visible console error.

## Additional context

- File: `store/audio.ts` — affected call sites: `playVerse` (66-68), `playGraph` (82-84), `next` (123-125), `prev` (134-136).
- Consumer that displays the desynced state: `components/audio/MiniPlayer.tsx:55,60`.
- Suggested fix: track a monotonically increasing "play generation" token, bump it on every play call, and have each `.then`/`.catch` check whether its captured token is still current before applying `set(...)`:
  ```ts
  let playToken = 0;

  function loadAndPlay(verse: AudioVerse, onEnded: () => void, token: number, set: ...) {
    const a = getAudio();
    a.onended = onEnded;
    a.src = getAudioUrl(verse.surah, verse.ayah);
    a.load();
    return a.play()
      .then(() => { if (token === playToken) set({ isLoading: false }); })
      .catch(() => { if (token === playToken) set({ isPlaying: false, isLoading: false }); });
  }

  // in each play entry point:
  const token = ++playToken;
  loadAndPlay(verse, () => get()._onEnded(), token, set);
  ```
  This preserves the existing behavior for the common case (no race) while making stale callbacks a no-op once a newer play call has superseded them.

---

Mubashir78:
## Description

Adds a monotonically increasing `playGen` counter to the audio store. Each play call captures the current token. Stale `.then`/`.catch` callbacks from superseded requests check `token === playGen` before applying state updates.

### The bug

`store/audio.ts` uses a single module-level `HTMLAudioElement` and reassigns `src` + calls `.load()` on every track change. If the user triggers two play calls in quick succession (e.g. double-tapping Next, or tapping a verse while another is loading), the browser rejects the \*first\* `play()` promise with `AbortError`. That stale rejection's `.catch()` then unconditionally sets `isPlaying: false`, even though the second track's `play()` has since resolved and audio is audibly playing.

### The fix

Every play entry point now increments `playGen` and captures the new value as `token`. The `.then`/`.catch` handlers only call `set(...)` when their captured token matches the current `playGen`, making stale callbacks a no-op.

### Affected call sites

- `playVerse`
- `playGraph`
- `next`
- `prev`

Fixes #102

### DIFF
--- a/store/audio.ts
+++ b/store/audio.ts
@@ -71,8 +71,12 @@ export const useAudioStore = create<AudioStore>((set, get) => ({
       queueIndex: 0,
     });
     loadAndPlay(verse, () => get()._onEnded())
-      .then(() => { if (token === playGen) set({ isLoading: false }); })
-      .catch(() => { if (token === playGen) set({ isPlaying: false, isLoading: false }); });
+      .then(() => {
+        if (token === playGen) set({ isLoading: false });
+      })
+      .catch(() => {
+        if (token === playGen) set({ isPlaying: false, isLoading: false });
+      });
   },
 
   playGraph: (verses) => {
@@ -88,8 +92,12 @@ export const useAudioStore = create<AudioStore>((set, get) => ({
       queueIndex: 0,
     });
     loadAndPlay(first, () => get()._onEnded())
-      .then(() => { if (token === playGen) set({ isLoading: false }); })
-      .catch(() => { if (token === playGen) set({ isPlaying: false, isLoading: false }); });
+      .then(() => {
+        if (token === playGen) set({ isLoading: false });
+      })
+      .catch(() => {
+        if (token === playGen) set({ isPlaying: false, isLoading: false });
+      });
   },
 
   pause: () => {
@@ -136,8 +144,12 @@ export const useAudioStore = create<AudioStore>((set, get) => ({
       isLoading: true,
     });
     loadAndPlay(verse, () => get()._onEnded())
-      .then(() => { if (token === playGen) set({ isLoading: false }); })
-      .catch(() => { if (token === playGen) set({ isPlaying: false, isLoading: false }); });
+      .then(() => {
+        if (token === playGen) set({ isLoading: false });
+      })
+      .catch(() => {
+        if (token === playGen) set({ isPlaying: false, isLoading: false });
+      });
   },
 
   prev: () => {
@@ -153,8 +165,12 @@ export const useAudioStore = create<AudioStore>((set, get) => ({
       isLoading: true,
     });
     loadAndPlay(verse, () => get()._onEnded())
-      .then(() => { if (token === playGen) set({ isLoading: false }); })
-      .catch(() => { if (token === playGen) set({ isPlaying: false, isLoading: false }); });
+      .then(() => {
+        if (token === playGen) set({ isLoading: false });
+      })
+      .catch(() => {
+        if (token === playGen) set({ isPlaying: false, isLoading: false });
+      });
   },
 
   _onEnded: () => {
--- a/store/audio.ts
+++ b/store/audio.ts
@@ -38,6 +38,12 @@ function getAudio(): HTMLAudioElement {
   return _audio!;
 }
 
+// Monotonically increasing generation counter. Each play call captures the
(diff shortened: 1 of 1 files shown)


nazarli-shabnam:
Reviewed this against #102.

**Fix: matches the issue exactly.** A monotonically increasing `playGen` token is bumped at every play entry point (`playVerse`, `playGraph`, `next`, `prev`), captured locally, and checked in both the `.then` and `.catch` handlers before applying `set(...)` — which is precisely the guard the issue proposed, and it correctly covers all four affected call sites with no gaps.

**CI:** `Lint · Type-check · Test` is failing, but again only at the `format:check` (Prettier) step, isolated to `store/audio.ts`. Typecheck/test/build never actually ran because format:check gates them — I don't see evidence of any other failure hiding behind it.

So: no functional/logic gap versus #102, this is a complete and correct fix. It's just blocked on CI being red due to formatting. Could you run `bun run format` on `store/audio.ts` and push? Once CI is green I think this is ready.

Mubashir78:
Thanks @nazarli-shabnam! Ran `prettier --write store/audio.ts` and pushed.

CI should go green now

nazarli-shabnam (approved):

PR #242, opened 2026-07-14
chore: add centralized environment variable validation at startup (#174)

### LINKED ISSUE #174: chore: no centralized environment variable validation at startup
## Problem
Required environment variables (`DATABASE_URL`, `QF_CLIENT_SECRET`, AI provider keys such as `ANTHROPIC_API_KEY`/`GEMINI_API_KEY`, etc.) are read ad hoc wherever they're needed across the codebase, with no single validation step at boot.

## Proposed fix
Add a small `lib/env.ts` (or similar) module that validates all required env vars once at startup (e.g. via a schema library already in the dependency tree, or a plain assertion module) and throws a clear, actionable error immediately if one is missing or malformed.

## Why this matters
Today a missing/misconfigured env var surfaces as an opaque runtime failure deep inside a request handler, rather than a clear boot-time error — this makes misconfiguration in new environments (local dev, staging, a new deploy target) much harder to diagnose than it needs to be.

Mubashir78:
### Link to issue number:
Closes #174

### Summary of the issue:
Required environment variables (DATABASE_URL, QF_CLIENT_SECRET, AI provider keys, etc.) are read ad hoc wherever they're needed across the codebase, with no single validation step at boot. A missing or misconfigured env var surfaces as an opaque runtime failure deep inside a request handler.

### Description of how this pull request fixes the issue:
Adds `lib/env.ts` with centralized validation that runs at module load time (imported in the root layout).

Key design decisions:
- **Required at boot:** NEXT_PUBLIC_APP_URL, NEXT_PUBLIC_QF_CLIENT_ID, NEXT_PUBLIC_QF_AUTH_BASE, QF_CLIENT_SECRET, QF_AUTH_BASE
- **Production-only required:** DATABASE_URL (dev has a fallback placeholder)
- **Optional:** REDIS_URL (app already handles missing Redis gracefully)
- **Deferred validation:** AI provider keys (ANTHROPIC_API_KEY / GEMINI_API_KEY) are validated when `getAiProvider()` is called, not at boot, so they're not needed for non-AI pages
- **Test-safe:** Validation is skipped when NODE_ENV=test or VITEST is set
- Exports typed accessor functions for every env var so callers get consistent defaults

### Testing:
- Verified module loads without error when required vars are set
- Verified it throws a clear error listing all missing vars at once
- Existing tests should be unaffected (test guard + vitest.setup.ts provides values)

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -1,3 +1,4 @@
+import "@/lib/env";
 import type { Metadata } from "next";
 import { Geist, Geist_Mono, Amiri } from "next/font/google";
 import { GoogleAnalytics } from "@next/third-parties/google";
--- a/lib/env.ts
+++ b/lib/env.ts
@@ -79,9 +79,7 @@ export function getAdminQfIds(): string[] {
 export function getAiProvider(): "claude" | "gemini" {
   const provider = env("AI_PROVIDER", "claude") as "claude" | "gemini";
   if (provider !== "claude" && provider !== "gemini") {
-    throw new Error(
-      `AI_PROVIDER must be "claude" or "gemini", got "${provider}"`
-    );
+    throw new Error(`AI_PROVIDER must be "claude" or "gemini", got "${provider}"`);
   }
   if (provider === "claude") {
     requireEnv("ANTHROPIC_API_KEY");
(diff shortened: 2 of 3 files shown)


nazarli-shabnam (changes requested):
lib/env.ts and the layout wiring are solid for #174 — required-at-boot list, deferred AI key validation, test-mode skip all match what the issue asked for.
Thanks for your contribution!

nazarli-shabnam (approved):

PR #243, opened 2026-07-14
fix: add rate limiting to bookmarks POST endpoint (#170)

### LINKED ISSUE #170: fix: /api/bookmarks POST has no rate limiting
## Problem
`app/api/bookmarks/route.ts` POST never calls the existing `rateLimitOrNull` helper from `lib/rate-limit.ts`. Every sibling mutation route (notes, workspace, challenges) rate-limits its POST handler; bookmarks is the outlier.

## Proposed fix
Add the same `rateLimitOrNull` check used in `app/api/notes/route.ts` / `app/api/workspace/route.ts` to the bookmarks POST handler, with limits consistent with the other per-user mutation routes.

## Why this matters
An authenticated user (or a compromised token) can spam bookmark inserts with no throttling, unlike every other mutation endpoint in the app.

Mubashir78:
Closes #170

Adds rateLimitOrNull check to the bookmarks POST handler, matching the pattern used by other mutation routes (challenges, notes, friends, activity, share).

### DIFF
--- a/app/api/bookmarks/route.ts
+++ b/app/api/bookmarks/route.ts
@@ -5,6 +5,7 @@ import { bookmarks } from "@/lib/infra/db/schema";
 import { requireUser } from "@/lib/auth/social-auth";
 import { isValidRef } from "@/lib/quran/quran-corpus";
 import { jsonError, parseJson } from "@/lib/infra/http";
+import { rateLimitOrNull } from "@/lib/infra/rate-limit";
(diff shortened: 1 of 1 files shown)


nazarli-shabnam:
Simple and consistent with how notes/workspace already do it. Thanks for closing this gap!

nazarli-shabnam (approved):

PR #244, opened 2026-07-14
fix: validate embedding response shape to prevent silent misalignment (#100)

### LINKED ISSUE #100: fix: embedBatch() doesn't validate response shape, risking silently misaligned embeddings
## Describe the bug

`embedBatch()` in `lib/ai.ts` sends a batch embedding request to Gemini's `batchEmbedContents` endpoint and returns the results with no validation that the response actually corresponds 1:1 with the input:

```ts
// lib/ai.ts, lines 56-93
async function embedViaRest(texts: string[]): Promise<number[][]> {
  ...
  const res = await fetch(
    `${EMBED_API_BASE}/models/${model}:batchEmbedContents?key=${apiKey}`,
    { method: "POST", ... body: JSON.stringify({ requests: texts.map((text) => ({ ... })) }) }
  );
  if (!res.ok) {
    const detail = await res.text().catch(() => "");
    throw new Error(`Embedding request failed: ${res.status} ${detail}`);
  }
  const data = (await res.json()) as { embeddings?: Array<{ values: number[] }> };
  return (data.embeddings ?? []).map((e) => e.values);   // <-- no length/shape check
}

/** Embeds many texts in one request, preserving input order. */
export async function embedBatch(texts: string[]): Promise<number[][]> {
  if (texts.length === 0) return [];
  return embedViaRest(texts);
}
```

The function's doc comment explicitly promises "preserving input order," and every caller relies on that guarantee by zipping the returned vectors back to their source texts **by array index** rather than by any explicit key. `scripts/embed-corpus.mjs` mirrors this same REST call and the same positional-zip pattern when seeding the `verse_embeddings` table (per the comment at `lib/ai.ts:41-47`: "The query and corpus must use the same model, so scripts/embed-corpus.mjs mirrors this exactly").

On a `200 OK` response, there is currently no check that:
- `data.embeddings.length === texts.length`
- every entry has a `values` array of the expected dimensionality (`EMBEDDING_DIMENSIONS = 768`)

If Gemini's API ever returns a partial batch (e.g. fewer embeddings than requests — which batch embedding APIs can do on partial per-item failures within an otherwise-`200` batch response) or an entry missing `values`, this function silently returns a shorter array, or an array containing `undefined` in place of a vector. Any caller doing positional zipping (e.g. the corpus-seeding script) would then attribute an embedding to the *wrong* verse reference for every entry after the misalignment point, or attempt to store `undefined` as a vector.

This is a data-integrity bug rather than a crash: nothing throws, so nothing alerts an operator. The corruption would only surface later as unexplained bad results in "search by meaning" / "find similar verses" (`lib/semantic-search.ts`, which reads directly from `verse_embeddings`), which is very hard to trace back to a batch-embedding response mismatch after the fact.

Note: the single-text path `embed()` (line 82-87) *does* guard against this — `if (!vector) throw new Error("No embedding returned")` — so this is specifically a gap in the batch path, not a project-wide gap in this file.

## Steps to reproduce

1. Trigger a corpus re-embed via `scripts/embed-corpus.mjs` (or any caller of `embedBatch`) while the Gemini `batchEmbedContents` endpoint returns a response where `embeddings.length < requests.length`, or where one entry lacks a `values` field.
   - This is hard to force deliberately without mocking the Gemini API, since it depends on upstream partial-failure behavior; the bug is confirmed by code inspection (no validation exists), not by having observed a live misalignment in production.
2. Observe that `embedBatch` returns fewer/malformed vectors than requested with no thrown error.
3. If a caller stores these results by position (as `scripts/embed-corpus.mjs` does), inspect `verse_embeddings` afterward for a ref/vector mismatch.

**Expected behaviour:** A batch response whose shape doesn't match the request should cause `embedBatch` to throw (matching the existing convention in `embed()`), so the caller's retry/error-handling logic engages instead of silently persisting misaligned data.

**Actual behaviour:** A malformed/partial batch response is returned as-is with no error, silently risking misaligned embeddings if consumed positionally.

## Environment

- Browser / OS: N/A — server-side/script-side bug
- Node.js version: N/A
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (requires a malformed upstream Gemini response to trigger; verified by code inspection)

N/A — no error is currently produced, which is the bug.

## Additional context

- File: `lib/ai.ts`, `embedViaRest` return statement, line 79.
- Suggested fix — validate shape before returning, mirroring the existing throw-on-empty pattern in `embed()`:
  ```ts
  const data = (await res.json()) as { embeddings?: Array<{ values: number[] }> };
  const out = (data.embeddings ?? []).map((e) => e.values);
  if (out.length !== texts.length || out.some((v) => !Array.isArray(v) || v.length !== EMBEDDING_DIMENSIONS)) {
    throw new Error(
      `Embedding batch mismatch: requested ${texts.length}, got ${out.length} well-formed vectors`
    );
  }
  return out;
  ```
- Related: `lib/semantic-search.ts` is the primary consumer of `verse_embeddings` and would be the place bad data eventually surfaces as wrong search/similarity results, which is why this is worth catching at the source rather than downstream.

---

Mubashir78:
Closes #100

The embedViaRest function previously cast the API response without validation. If Gemini returned missing or malformed embeddings (wrong count or wrong dimensionality), the error would propagate as a cryptic downstream failure or, worse, silently store misaligned vectors.

Changes:
- Verify the response contains exactly the expected number of embeddings
- Verify each embedding has exactly 768 values (EMBEDDING_DIMENSIONS)
- Throw a clear error with the expected vs actual shape for either case

### DIFF
--- a/__tests__/lib/ai/ai.test.ts
+++ b/__tests__/lib/ai/ai.test.ts
@@ -11,6 +11,12 @@ function jsonResponse(json: unknown, ok = true, status = 200) {
   } as Response;
 }
 
+function vec768(...seed: number[]): number[] {
+  const v = Array<number>(EMBEDDING_DIMENSIONS).fill(0);
+  for (let i = 0; i < seed.length && i < v.length; i++) v[i] = seed[i];
+  return v;
+}
+
 describe("embeddings", () => {
   beforeEach(() => {
     process.env.GEMINI_API_KEY = "test-key";
@@ -22,13 +28,14 @@ describe("embeddings", () => {
   });
 
   it("embed() returns the vector for a single text and requests 768 dims", async () => {
+    const expected = vec768(0.1, 0.2, 0.3);
     const fetchMock = vi
       .fn()
-      .mockResolvedValue(jsonResponse({ embeddings: [{ values: [0.1, 0.2, 0.3] }] }));
+      .mockResolvedValue(jsonResponse({ embeddings: [{ values: expected }] }));
     vi.stubGlobal("fetch", fetchMock);
 
     const v = await embed("patience");
-    expect(v).toEqual([0.1, 0.2, 0.3]);
+    expect(v).toEqual(expected);
 
     const body = JSON.parse((fetchMock.mock.calls[0][1] as RequestInit).body as string);
     expect(body.requests[0].content.parts[0].text).toBe("patience");
@@ -36,16 +43,19 @@ describe("embeddings", () => {
   });
 
   it("embedBatch() preserves input order", async () => {
+    const a = vec768(1);
+    const b = vec768(2);
+    const c = vec768(3);
     vi.stubGlobal(
       "fetch",
       vi
         .fn()
         .mockResolvedValue(
-          jsonResponse({ embeddings: [{ values: [1] }, { values: [2] }, { values: [3] }] })
+          jsonResponse({ embeddings: [{ values: a }, { values: b }, { values: c }] })
         )
     );
     const out = await embedBatch(["a", "b", "c"]);
-    expect(out).toEqual([[1], [2], [3]]);
+    expect(out).toEqual([a, b, c]);
   });
 
   it("embedBatch() short-circuits on empty input without calling the API", async () => {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
This closes #100 exactly right — checking both the array length and each vector's dimensionality before returning, matching the throw-on-empty pattern embed() already had. Nice touch bumping the test mocks to real 768-dim vectors too. Thanks!

nazarli-shabnam (approved):

PR #245, opened 2026-07-14
chore: remove unused @tanstack/react-query dependency (#180)

### LINKED ISSUE #180: chore: @tanstack/react-query installed but unused; inconsistent ad-hoc data-fetching patterns
## Problem
`@tanstack/react-query`'s `QueryClientProvider` is wired up in `components/providers.tsx`, but no component anywhere actually uses `useQuery`/`useMutation`/`useInfiniteQuery` — the provider wraps the app for nothing. Instead, every feature hand-rolls its own fetch/cache/loading-state plumbing independently: `app/social/page.tsx` uses raw `fetch` + `useState`, `components/admin/*` uses a custom `useAsync` hook (`components/admin/useAsync.ts`), and some server-derived state (streak, pending counts) is fetched ad hoc from multiple components (e.g. both `app/social/page.tsx` and `components/social/StreakBadge.tsx` independently fetch overlapping social data).

## Proposed fix
Either adopt `react-query` consistently as the shared data-fetching/caching layer across the app (unifying retries, loading/error states, and de-duplicating requests for the same data) now that the provider is already in place, or remove the unused provider/dependency if it's not going to be adopted.

## Why this matters
As-is, the provider is dead weight that does nothing, and the current ad-hoc pattern means near-identical loading/error/retry logic is reimplemented per feature, with the same server data sometimes fetched redundantly by multiple components on the same page.

---
*Edit: corrected — the QueryClientProvider is present (`components/providers.tsx`), just unused by any consumer. Original wording ("installed but unused anywhere") was imprecise.*

Mubashir78:
Closes #180

- @tanstack/react-query was installed but never used, zero useQuery/useMutation calls across the codebase.
- Removed the dependency, the QueryClient setup, and
the QueryClientProvider wrapper from the root Providers component.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -14,7 +14,6 @@
         "@radix-ui/react-separator": "^1.1.8",
         "@radix-ui/react-slot": "^1.2.4",
         "@radix-ui/react-tooltip": "^1.2.8",
-        "@tanstack/react-query": "^5.101.2",
         "@xyflow/react": "^12.10.2",
         "axios": "^1.16.1",
         "class-variance-authority": "^0.7.1",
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
Confirmed — zero useQuery/useMutation calls anywhere, so this was genuinely dead weight. Clean removal, nothing left behind. Thanks!

PR #247, opened 2026-07-14
fix: reduce self-heal queries and batch score computation on challenges GET (#176, #103)

### LINKED ISSUE #103: fix: N+1 query storm recomputing scores for every completed challenge on each load
## Describe the bug

`GET /api/social/challenges` (`app/api/social/challenges/route.ts`) recomputes each challenge's score on nearly every load, rather than reusing a persisted score for challenges that are already finished:

```ts
// app/api/social/challenges/route.ts, lines 79-96
const enriched = await Promise.all(
  rows.map(async (c) => {
    const needsScores = c.status === "active" || c.status === "completed";
    const cached = resolvedScores.get(c.id);
    const [challengerScore, challengedScore] = cached
      ? [cached.challengerScore, cached.challengedScore]
      : needsScores
      ? await Promise.all([scoreChallenge(c.challengerId, c), scoreChallenge(c.challengedId, c)])
      : [0, 0];
    return { ...c, ..., challengerScore, challengedScore };
  })
);
```

`resolvedScores` (populated earlier in the handler, not shown above) only holds scores for challenges that transitioned from `active` to `completed` *during this specific request* (i.e. just-resolved ones). Every other row with `status === "active"` or `status === "completed"` — which, for a completed challenge, is a score that can never change again — falls through to `scoreChallenge(...)`, which issues DB `COUNT` queries per user per challenge.

The handler loads up to 200 challenges (`.limit(200)` a few lines earlier), so a user with a long challenge history can trigger up to ~200 × 2 = 400 `scoreChallenge` calls per page load, each running its own `COUNT` query, all fired concurrently via `Promise.all`. `lib/db.ts:14` caps the Postgres connection pool at `max: 10`. A burst of ~400 concurrent queries against a 10-connection pool means most of them queue rather than run in parallel, so the endpoint's latency scales with the user's total challenge count instead of being roughly constant, and a spike in traffic to this endpoint (several such users loading the tab around the same time) can saturate the shared pool for other requests too.

The wasted work is entirely avoidable for `completed` challenges specifically: a completed challenge's `winnerId` is already persisted (implied by the `status === "completed"` transition logic elsewhere in this file), so its final scores are, by definition, never going to change again — recomputing them on every subsequent load has no purpose beyond the one time the challenge was resolved.

## Steps to reproduce

1. As a test user, accumulate a large number of completed challenges (or seed the `challenges` table directly with, say, 100+ rows with `status = 'completed'` for one user).
2. Call `GET /api/social/challenges` as that user.
3. Observe the number of DB queries issued (e.g. via Postgres query logging, or by timing the request) relative to the number of completed challenges.

**Expected behaviour:** Completed challenges' scores are read from a persisted value (or otherwise computed once, at resolution time) rather than recomputed on every subsequent page load — request latency and DB load should not scale linearly with a user's total historical challenge count.

**Actual behaviour:** Every `active` or `completed` challenge not resolved in the current request re-triggers two `COUNT` queries, so DB load and latency scale with total challenge count, and a heavy user can generate hundreds of concurrent queries against a 10-connection pool from a single page load.

## Environment

- Browser / OS: N/A — server-side/DB performance issue
- Node.js version: N/A
- Is this reproducible on openhikmah.com? [ ] Yes  [ ] No  [x] Not checked (verified by code inspection and query-count reasoning; not load-tested against the live deployment)

N/A — this is a latency/load characteristic, not an error.

## Additional context

- File: `app/api/social/challenges/route.ts`, lines 79-96 (the `enriched` computation). Query limit is set at line 65 (`.limit(200)`), pool size at `lib/db.ts:14` (`max: 10`).
- Related, milder instance of the same shape: `lib/graph-service.ts:43` (`hydrate`) resolves each edge's verse via a per-row `resolveVerse` call inside `Promise.all(rows.map(...))` instead of the batch `getVerses` helper already used elsewhere in the codebase (e.g. `lib/connection-generator.ts:112`) — bounded and parallel so lower severity, but the same avoidable N+1 pattern; worth fixing alongside this if convenient, though it's a separate code path.
- Suggested fix: persist `challengerScore`/`challengedScore` on the `challenges` row at the moment a challenge transitions to `completed` (alongside the existing `winnerId` write), then read those persisted columns directly for any row where `status === "completed"`, reserving the live `scoreChallenge` computation for `status === "active"` rows only (which is a much smaller, bounded set — a user typically has few simultaneously active challenges).

---

### LINKED ISSUE #176: fix: challenges GET performs redundant self-heal writes on every request
## Problem
`app/api/social/challenges/route.ts` GET performs two extra DB read queries plus a resolve/"self-heal" write step on every single page load before returning the list, with no debounce, lock, or last-run timestamp to skip redundant work.

## Proposed fix
Add a lightweight guard (e.g. a last-run timestamp per user, or a short-lived lock) so the self-heal step only actually executes when it's likely to be needed, instead of on every GET.

## Why this matters
A chatty client (polling, rapid navigation) or a bot triggers redundant self-heal writes on every single request, adding unnecessary DB load for no benefit beyond the first call in a session.

Mubashir78:
Closes #176
Closes #103

Two changes to the challenges GET handler:

1. **#176**: Combined two separate self-heal SELECT queries into one using
   `inArray(['active', 'pending'])`, then dispatches to the appropriate
   resolver based on status. Halves the self-heal query overhead per request.

2. **#103**: Replaced the `rows.map` with unbounded `Promise.all`
   (fire N×2 concurrent score queries) with `mapWithConcurrency` at 8-wide
   concurrency, after filtering out already-cached scores from the self-heal
   step. Eliminates the N+1 query storm for users with many challenges.

### DIFF
--- a/__tests__/api/social/challenges.test.ts
+++ b/__tests__/api/social/challenges.test.ts
@@ -299,15 +299,13 @@ describe("GET /api/social/challenges", () => {
   });
 
   it("bounds the challenges query with orderBy + limit(200)", async () => {
-    // NOTE: the route issues the unbounded active-resolution select, then the
-    // unbounded pending-expiry select, then the capped display select. The
-    // recording chain is wired to that third select. If the route is
-    // reordered, update which Once() carries the recording chain.
+    // NOTE: the route issues the unbounded combined resolution select (active or
+    // pending that has ended), then the capped display select. The recording
+    // chain is wired to the second select.
     authedAs(makeUser({ id: 1 }));
     const calls: Record<string, unknown[][]> = {};
     mockSelect
-      .mockReturnValueOnce(makeDbChain([])) // resolution query (no ended-active)
-      .mockReturnValueOnce(makeDbChain([])) // pending-expiry query (none expired)
+      .mockReturnValueOnce(makeDbChain([])) // combined resolution (none expired)
       .mockReturnValueOnce(makeRecordingChain([], calls)); // capped display query
     const res = await GET(makeGetReq());
     expect(res.status).toBe(200);
@@ -326,8 +324,7 @@ describe("GET /api/social/challenges", () => {
     authedAs(makeUser({ id: 1 }));
     const challenge = makeChallenge({ status: "pending" });
     mockSelect
-      .mockReturnValueOnce(makeDbChain([])) // resolution query (none ended-active)
-      .mockReturnValueOnce(makeDbChain([])) // pending-expiry query (none expired)
+      .mockReturnValueOnce(makeDbChain([])) // combined resolution (none expired)
       .mockReturnValueOnce(makeDbChain([challenge])) // capped display query
       .mockReturnValueOnce(
         makeDbChain([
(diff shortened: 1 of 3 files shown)


Mubashir78:
I can split the two fixes into two separate PRs if required as per guidelines.

nazarli-shabnam:
Nice fix — this is exactly what #176 and #103 needed. Folding the two self-heal selects into one with inArray, then bounding the score computation with mapWithConcurrency instead of firing N×2 unbounded queries, is clean and the tests track the new query shape. Thanks for this!

nazarli-shabnam (approved):

PR #248, opened 2026-07-15
fix: add rate limiting and query-length cap to keyword search (#173)

### LINKED ISSUE #173: fix: keyword search endpoint has no rate limiting or query-length cap
## Problem
`app/api/search/route.ts` GET has two branches: the "meaning" (AI) branch is throttled via `consume()`, but the default keyword-search branch is unauthenticated and completely unrate-limited, with no cap on query length before the query is forwarded to `api.quran.com`.

## Proposed fix
Apply the same rate-limiting utility (`lib/rate-limit.ts`) used elsewhere to the keyword-search branch, and add a reasonable max length on the incoming query string.

## Why this matters
This is a public, unauthenticated endpoint — as written, an attacker can proxy unlimited requests through it to an external upstream API (api.quran.com), which is both a cost/availability risk for that upstream and a potential abuse vector against this app's own infrastructure.

Mubashir78:
Fixes #173

### Summary of the issue
The keyword search endpoint had no rate limiting and no query-length
cap, allowing unbounded requests with arbitrarily long query strings.

### Description of user facing changes
- Keyword search now returns 429 Too Many Requests when the per-IP
  rate limit is exceeded
- Queries longer than 200 characters return 400 Bad Request

### Description of developer facing changes
None.

### Description of development approach
Added a `consume()` call before the keyword search execution (matching
the existing semantic search rate limit), and a `MAX_QUERY_LENGTH`
constant (200) validated early in the handler.

### Testing strategy
1. Send a query with 201+ characters -- expect 400
2. Send rapid keyword search requests from the same IP -- expect 429
   after the rate limit is exceeded
3. Normal keyword and semantic search should behave unchanged
4. Verify existing tests still pass with `bun run test:ci`

### Known issues with pull request
None.

### DIFF
--- a/__tests__/api/search.test.ts
+++ b/__tests__/api/search.test.ts
@@ -274,15 +274,15 @@ describe("GET /api/search", () => {
     expect((await res.json()).results[0].ref).toBe("2:1");
   });
 
-  it("does not rate-limit the keyword search response itself", async () => {
+  it("rate-limits the keyword search response when the search budget is exhausted", async () => {
     mockFetch.mockResolvedValueOnce(
       quranComResponse([{ verse_key: "1:1", translations: [{ text: "In the name of God" }] }])
     );
-    mockConsume.mockResolvedValue(false); // search-log budget exhausted
+    mockConsume.mockResolvedValue(false);
     const res = await GET(makeSearchReq("mercy"));
-    expect(res.status).toBe(200);
+    expect(res.status).toBe(429);
     const body = await res.json();
-    expect(body.results[0].ref).toBe("1:1");
+    expect(body.error).toBe("Too many search requests");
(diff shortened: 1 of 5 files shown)


nazarli-shabnam (changes requested):
The actual fix for #173 looks right — 429 when the rate limit is exhausted, 400 over 200 chars, keyword path unchanged otherwise. But this branch is stacked on top of #245's (still-unmerged) commits, so the diff here also drags in the react-query removal — bun.lock, providers.tsx, package.json — which has nothing to do with this issue.
I am merging this time, as its not a big deal. Consider to create branch off main next time.

nazarli-shabnam (approved):

Issue #185, opened 2026-07-10
Fetch more (not same) results on repeat theme/root/contrast expansion


nazarli-shabnam:
## Context
Follow-up from #87 / PR #184. That PR handles duplicate-verse highlighting and new-node orientation on the canvas. A third idea from the same discussion — repeat-clicking the same expansion kind (theme/root/contrast) on a verse should surface *additional* related verses instead of the same cached/regenerated set — needs backend changes, so it's scoped out here.

## Problem
Clicking `+` → theme (or root/contrast) on a verse that's already had that expansion run just returns the same connections again (cache hit) or regenerates a similar first-N set (cache miss). There's no way to ask for "more."

## Proposed change
- Add `excludeRefs: string[]` to the `/api/connections` POST body (`app/api/connections/route.ts`).
- `getConnections` (`lib/ai/graph-service.ts`) — when `excludeRefs` is present, the cache-hit DB query should exclude those refs (e.g. `notInArray`) instead of always returning the same top-200.
- `discoverCandidates` (`lib/ai/connection-discovery.ts`) — both `semanticCandidates` (via `nearest()` in `lib/quran/semantic-search.ts`) and `rootCandidates` currently only exclude the single source ref (`ne(...)`); both need to accept and apply a full exclude list (`notInArray(...)`) for the AI-generation miss path.
- Frontend (`HikmahCanvas.tsx`'s `runExpansion`, `ExpandMenu.tsx`) — when the same `kind` is clicked again for a node that already has expansion children of that kind, collect those children's refs (derivable from existing edges) and pass them as `excludeRefs`.

## Why not bundled with #87
Touches the API contract, DB query shape, and both candidate-discovery paths (semantic + root) — a meaningfully larger and riskier change than the pure-frontend duplicate/highlight work, worth its own review cycle.

nazarli-shabnam:
Resolved by #186 (merged).

PR #287, opened 2026-07-16
test: cover 5 untested lib modules flagged by the coverage gate

### LINKED ISSUE #268: chore: no CI coverage collection or thresholds
## Problem
\`test:ci\` in \`package.json\` runs \`vitest run\` without ever passing \`--coverage\`, so coverage isn't generated in CI at all. \`vitest.config.ts\`'s coverage block configures \`provider\`/\`reporter\`/\`include\`/\`exclude\` but has no \`thresholds\`, so even a local coverage run has nothing that would fail on a regression.

## Impact
Coverage can silently erode over time with no CI signal to catch it — including entire modules shipping with zero tests. As a concrete example, several non-trivial \`lib/**\` modules currently have no matching test file at all: \`lib/quran/verse-resolver.ts\` (verse resolution + live API fallback + surah bounds validation), \`lib/ai/connection-discovery.ts\` (root-candidate SQL aggregation logic), \`lib/ai/prompt-registry.ts\` (template rendering + cache), \`lib/canvas/share-canvas.ts\`'s \`isValidNode\` (a security-relevant guard rejecting malformed shared-canvas payloads before storage/rendering), and \`lib/admin/admin-audit.ts\`'s \`serializeMeta\` (BigInt/circular-reference handling). None of this is flagged today because there's no coverage gate in CI at all.

## Location
\`.github/workflows/ci.yml:41-42\` (\`test:ci\` → \`vitest run\`), \`vitest.config.ts:14-19\` (coverage block, no \`thresholds\`)

## Suggested fix
Add \`--coverage\` to \`test:ci\` (or a dedicated coverage CI step) and set \`coverage.thresholds\` in \`vitest.config.ts\` (globally or per-directory for \`lib/**\`) so a coverage drop fails the build. As a starting point, add focused unit tests for the specific untested modules called out above.

nazarli-shabnam:
## Problem
Closes #268 (remaining scope). The CI coverage gate itself (`test:coverage` + `vitest.config.ts` thresholds) was already added in #286. This PR fills in the specific untested modules the issue called out, all of which had zero test coverage:

- `lib/quran/verse-resolver.ts` — verse resolution + live API fallback + surah bounds validation
- `lib/ai/connection-discovery.ts` — root-candidate SQL aggregation logic
- `lib/ai/prompt-registry.ts` — template rendering + cache
- `lib/canvas/share-canvas.ts`'s `isValidNode` — security-relevant guard rejecting malformed shared-canvas payloads
- `lib/admin/admin-audit.ts`'s `serializeMeta` (via `logAdminAction`) — BigInt/circular-reference handling

## Fix
Added focused unit tests for each module, following the existing db-mocking conventions in `__tests__/lib/ai/connection-generator.test.ts` and `__tests__/lib/quran/verse-of-day.test.ts` (mocked `db` select chains, no real Postgres connection needed).

## Test plan
- `bun run lint -- --max-warnings=0` — clean
- `bun run typecheck` — clean
- `bun run test:coverage` — 651/651 tests pass; statements 67.81%, branches 69.55%, functions 64.51%, lines 69.32% (all above the 65/65/60/65 floor)

### DIFF
--- a/__tests__/lib/admin/admin-audit.test.ts
+++ b/__tests__/lib/admin/admin-audit.test.ts
@@ -1,0 +1,76 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const { mockValues, mockInsert, mockIncr } = vi.hoisted(() => {
+  const mockValues = vi.fn().mockResolvedValue(undefined);
+  return {
+    mockValues,
+    mockInsert: vi.fn(() => ({ values: mockValues })),
+    mockIncr: vi.fn(),
+  };
+});
+
+vi.mock("@/lib/infra/db", () => ({ db: { insert: mockInsert } }));
+vi.mock("@/lib/infra/metrics", () => ({ incr: mockIncr }));
+
+import { logAdminAction } from "@/lib/admin/admin-audit";
+
+describe("logAdminAction", () => {
+  beforeEach(() => {
+    mockValues.mockReset();
+    mockValues.mockResolvedValue(undefined);
+    mockInsert.mockClear();
+    mockIncr.mockClear();
+  });
+
+  it("inserts a row with meta JSON-stringified normally", async () => {
+    await logAdminAction({
+      adminQfId: "qf-1",
+      action: "flag.delete",
+      meta: { key: "ai_gen_limit" },
+    });
+    expect(mockValues).toHaveBeenCalledWith(
+      expect.objectContaining({
+        adminQfId: "qf-1",
+        action: "flag.delete",
+        meta: JSON.stringify({ key: "ai_gen_limit" }),
+      })
+    );
+  });
+
+  it("stores null meta when meta is undefined", async () => {
+    await logAdminAction({ adminQfId: "qf-1", action: "noop" });
+    expect(mockValues).toHaveBeenCalledWith(expect.objectContaining({ meta: null }));
+  });
+
+  it("defaults targetType/targetId to null when omitted", async () => {
+    await logAdminAction({ adminQfId: "qf-1", action: "noop" });
+    expect(mockValues).toHaveBeenCalledWith(
+      expect.objectContaining({ targetType: null, targetId: null })
+    );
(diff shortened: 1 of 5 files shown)


nazarli-shabnam:
### Code review

Ran a multi-agent review pass (CLAUDE.md compliance, shallow bug scan, git-history cross-check, past-PR comment check, source-comment compliance check). Two independent agents (bug scan + comment-compliance) converged on the same issue, and the git-history agent surfaced a real coverage gap tied to a past regression fix. Both addressed in 5392c15:

1. Misleading "prototype pollution" test in `share-canvas.test.ts` — `JSON.parse` sets `__proto__` as an own data property, not the object's real prototype, so the test passed for the same reason the adjacent "missing surahName/translation" test does, not because of any pollution-specific guard. Replaced with a test that verifies what actually happens (extra fields are ignored, no real prototype is touched).

https://github.com/OpenHikmah/openhikmah-web/blob/5392c15c25e428c0daa6d49c4b290ff688272864/__tests__/lib/canvas/share-canvas.test.ts#L53-L61

2. No regression coverage for the `excludeRefs` → `notInArray` wiring in `rootCandidates()`, added in 2297aab to fix #185 (repeat-click expansion returning stale refs). The original "root" test never passed `excludeRefs`, so the exclude-list branch was untested. Added two tests that inspect the actual `and(...)` call shape (verified they fail if the wiring regresses — reverted the source locally to confirm before restoring it).

https://github.com/OpenHikmah/openhikmah-web/blob/5392c15c25e428c0daa6d49c4b290ff688272864/lib/ai/connection-discovery.ts#L40-L47
https://github.com/OpenHikmah/openhikmah-web/blob/5392c15c25e428c0daa6d49c4b290ff688272864/__tests__/lib/ai/connection-discovery.test.ts#L82-L102

No CLAUDE.md/AGENTS.md compliance issues, and no concerns carried over from past PR review comments on the touched source files.

PR #289, opened 2026-07-16
test: fix misleading proto-pollution test + excludeRefs regression coverage

### LINKED ISSUE #185: Fetch more (not same) results on repeat theme/root/contrast expansion

nazarli-shabnam:
## Context
Follow-up to #287 (already merged). #287's own \`/code-review:code-review\` pass surfaced two real issues after the PR had already been merged — this PR lands the fixes on top of \`main\`.

## Problem
1. \`__tests__/lib/canvas/share-canvas.test.ts\`'s "prototype pollution" test passed for the wrong reason: \`JSON.parse\` sets \`__proto__\` as a plain own data property, not the object's actual prototype, so the test passed only because the required \`surahName\`/\`translation\` fields were absent — the same reason an adjacent test already passes. It gave false confidence that a security property was covered.
2. \`__tests__/lib/ai/connection-discovery.test.ts\` had no coverage for the \`excludeRefs\` → \`notInArray\` wiring in \`rootCandidates()\` (added in 2297aab to fix #185 — repeat-click expansion returning stale refs). The one "root" test never passed \`excludeRefs\`, so a regression there would go undetected.

## Fix
- Replaced the misleading test with one that verifies what \`isValidNode\` actually does with an extra \`__proto__\`-named field (ignores it, doesn't touch the real prototype).
- Added two tests that inspect the actual \`and(...)\` where-clause call shape drizzle builds, to assert \`notInArray(excludeRefs)\` is included when \`excludeRefs\` is non-empty and omitted when it's empty. Verified these fail by temporarily reverting the source wiring locally, then restored it.

## Test plan
- \`bun run lint -- --max-warnings=0\` — clean
- \`bun run typecheck\` — clean
- \`bun run vitest run __tests__/lib/ai/connection-discovery.test.ts __tests__/lib/canvas/share-canvas.test.ts\` — 16/16 pass

### DIFF
--- a/__tests__/lib/ai/connection-discovery.test.ts
+++ b/__tests__/lib/ai/connection-discovery.test.ts
@@ -1,7 +1,11 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 
-// ── DB mock: select() resolves to a configurable result via a chainable Proxy ──
-function makeSelectChain(resolveWith: unknown[]) {
+// ── DB mock: select() resolves to a configurable result via a chainable Proxy.
+// Optionally records the arguments each chained method was called with, keyed
+// by method name, so a test can assert *which* where-clause conditions were
+// built (e.g. whether notInArray(excludeRefs) was included) without needing to
+// fully emulate drizzle's SQL builder.
+function makeSelectChain(resolveWith: unknown[], calls?: Record<string, unknown[][]>) {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const chain: any = new Proxy(
     function () {
@@ -12,7 +16,10 @@ function makeSelectChain(resolveWith: unknown[]) {
         if (prop === "then")
           return (res: (v: unknown) => unknown, rej?: (e: unknown) => unknown) =>
             Promise.resolve(resolveWith).then(res, rej);
-        return () => chain;
+        return (...args: unknown[]) => {
+          if (calls) (calls[prop as string] ??= []).push(args);
+          return chain;
+        };
       },
       apply() {
         return chain;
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance — the new `excludeRefs`/`notInArray` assertions match `lib/ai/connection-discovery.ts`'s actual conditional wiring, and the replacement `share-canvas.test.ts` assertions accurately reflect `isValidNode`'s behavior.

PR #293, opened 2026-07-16
feat: @mentions in verse notes with friends-only notifications

### LINKED ISSUE #117: feat: @mentions in verse notes with notifications
## Feature description

Let users tag other users in verse notes via `@username`, with a notification sent to the mentioned user — built entirely on the app's own database, not QF's OAuth-provided data.

## Problem it solves

Verse notes (`verse_notes` table) are currently a solo feature — there's no way to draw a friend's attention to a note on a specific verse, even though the app already has a working friends/social graph (streaks, friends, leaderboard, challenges) and an existing notification pattern (the friend-request pending-badge). @mentions would extend that existing social layer into the notes feature.

## Proposed solution

- Parse `@username` tokens when a note is saved, resolve against the `users` table (case-insensitive, matching the existing username lookup convention already used for friend search).
- On a valid mention, create a notification for the mentioned user — reusing the existing friend-badge notification pattern (already built for pending friend requests) rather than inventing a new notification mechanism.
- Mention resolution should probably be scoped to friends only by default (matching the existing social graph's trust boundary), rather than allowing @mentioning any username platform-wide — see "Open questions" below.
- This is explicitly **not** QF's "Tags" OAuth scope — that scope is for QF-side content labeling, unrelated to this in-app social feature. No new OAuth scope needed; this stays entirely within the app's own `users` + `verse_notes` tables.

## Theological considerations

None directly — this is a social/notification feature layered on user-authored notes, not a change to canonical text or AI-articulated connections. Worth keeping mention-triggered notifications lightweight and non-intrusive so they don't feel like noise on a reflective/study-focused surface.

## Alternatives considered

- **Mention visibility: friends-only vs. anyone.** Friends-only is more consistent with the existing social trust model (friend requests, streak comparisons) and avoids unsolicited-mention spam from strangers; anyone-can-mention is more open but has no precedent elsewhere in the app's social features. Flagging as an open decision rather than assuming — this was called out as undecided in the original roadmap note too.
- **Email/push notification instead of in-app badge:** the in-app badge pattern is what's already built and working for friend requests; extending that is lower-effort and more consistent than introducing a new notification channel for this one feature.

## Additional context

Originally scoped as new "future scope" during internal roadmap planning; filed here as a tracked issue. Relevant existing files: `lib/db/schema.ts` (`verse_notes`, `users`), the existing friend-request pending-badge notification UI (for the pattern to reuse).

nazarli-shabnam:
## Problem
Closes #117. Verse notes are currently a solo feature — no way to draw a friend's attention to a note on a specific verse, even though the app has a working friends/social graph and an existing notification-badge pattern (the friend-request pending badge).

## Design decisions
- **Friends-only, not platform-wide.** @mentions resolve only against the mentioning user's accepted friends, matching the existing social graph's trust boundary and avoiding unsolicited-mention spam from strangers.
- **New \`note_mentions\` table (migration \`0017_violet_gamma_corps.sql\`).** There's no existing generic notifications table — the friend-request badge is a live count derived from \`friendships\` rows (\`status = 'pending'\`), not backed by a durable per-user record. A mention needs its own durable row since there's no equivalent status transition to derive it from.
- **Not** QF's "Tags" OAuth scope — stays entirely within the app's own \`users\`/\`verse_notes\` tables, no new OAuth scope.

## Fix
- \`lib/social/mentions.ts\`: \`parseMentionedUsernames\` (pure regex extraction) + \`resolveFriendMentions\` (case-insensitive match via the existing \`lower(username) = lower(...)\` convention from \`app/api/social/friends/route.ts\`, scoped to accepted friendships).
- \`app/api/notes/route.ts\`'s \`POST\` now parses mentions out of the saved note and inserts a \`note_mentions\` row per resolved friend mention — best-effort, a mention failure never fails the note save.
- New \`GET\`/\`PATCH /api/social/mentions\` (list + unread count / mark-all-read), a new \`/mentions\` page to view them (with a "Open in canvas" link per mention, matching the Bookmarks page pattern), and a badge in \`AccountMenu\` (\`pendingMentionCount\` in \`store/social.ts\`, polled in \`Header.tsx\` every 60s alongside the existing friend-request poll).

## ⚠️ Action needed after merge
This PR adds a new table via a Drizzle migration (\`lib/infra/db/migrations/0017_violet_gamma_corps.sql\`). **After merging, run migrations against the deployed database** (the same step needed for any schema change in this repo):
\`\`\`
DATABASE_URL=<production DB URL> bun scripts/migrate.mjs
\`\`\`
No new environment variables or OAuth scopes are needed.

## Test plan
- \`bun run lint -- --max-warnings=0\` — clean
- \`bun run typecheck\` — clean
- Applied the migration directly against a real local Postgres (via \`docker compose\`) to verify it's valid SQL and the FK/index shape is correct, then rolled it back to leave the dev DB clean.
- New unit tests: \`__tests__/lib/social/mentions.test.ts\` (10 tests — parsing + friend-resolution logic), \`__tests__/api/social/mentions.test.ts\` (GET/PATCH), extended \`__tests__/api/notes.test.ts\` (mention insertion + best-effort failure handling).
- \`bun run test:coverage\` — 674/674 tests pass, coverage above the configured floor.

### DIFF
--- a/__tests__/api/notes.test.ts
+++ b/__tests__/api/notes.test.ts
@@ -25,16 +25,28 @@ function makeDbChain(resolveWith: unknown = []) {
   return chain;
 }
 
-const { mockSelect, mockInsert, mockRateLimitOrNull } = vi.hoisted(() => ({
-  mockSelect: vi.fn(() => makeDbChain([])),
-  mockInsert: vi.fn(() => makeDbChain([])),
-  mockRateLimitOrNull: vi.fn(async (): Promise<NextResponse | null> => null),
-}));
+const { mockSelect, mockInsert, mockRateLimitOrNull, mockResolveFriendMentions } = vi.hoisted(
+  () => ({
+    mockSelect: vi.fn(() => makeDbChain([])),
+    mockInsert: vi.fn(() => makeDbChain([])),
+    mockRateLimitOrNull: vi.fn(async (): Promise<NextResponse | null> => null),
+    mockResolveFriendMentions: vi.fn(async () => [] as Array<{ id: number; username: string }>),
+  })
+);
 
 vi.mock("@/lib/infra/db", () => ({ db: { select: mockSelect, insert: mockInsert } }));
 vi.mock("@/lib/infra/rate-limit", () => ({
   rateLimitOrNull: mockRateLimitOrNull,
 }));
+// Real parseMentionedUsernames (pure), mocked resolveFriendMentions (db-backed)
+// so mention tests don't need to mock the full friends/users query chain.
+vi.mock("@/lib/social/mentions", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/social/mentions")>();
+  return {
+    parseMentionedUsernames: actual.parseMentionedUsernames,
+    resolveFriendMentions: mockResolveFriendMentions,
+  };
+});
 
 import { GET, POST } from "@/app/api/notes/route";
 import { requireUser } from "@/lib/auth/social-auth";
@@ -109,6 +121,8 @@ describe("POST /api/notes", () => {
     mockInsert.mockReturnValue(makeDbChain([{ id: 1, verseRef: "2:255", note: "hi" }]));
     mockRateLimitOrNull.mockReset();
     mockRateLimitOrNull.mockResolvedValue(null);
+    mockResolveFriendMentions.mockReset();
+    mockResolveFriendMentions.mockResolvedValue([]);
   });
 
   it("401 when unauthenticated", async () => {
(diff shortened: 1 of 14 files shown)


nazarli-shabnam:
### Code review

No issues found. Ran a multi-agent review pass (CLAUDE.md/AGENTS.md compliance, shallow bug scan, git-history cross-check, past-PR comment check, source-comment compliance check) — SQL is parameterized via Drizzle's tagged \`sql\` template (no injection risk), friend-resolution is scoped correctly in both friendship directions, the notes-POST mention side-effect is properly best-effort (wrapped in its own try/catch, can't fail the note save), and the note_id ordering (insert note, then use its returned id for mentions) is correct.

One point came up across two of the agents worth noting but not blocking: \`PATCH /api/social/mentions\` has no \`rateLimitOrNull\` call, unlike other mutation routes in this codebase. Two other agents judged this acceptable since it's idempotent, takes no user-controlled input, and only touches rows already scoped to the caller — consistent with \`GET /api/social/friends\` also being unrated. Flagging for visibility, not requesting a change.

PR #303, opened 2026-07-26
chore: extract DB migrations to a one-shot deploy step

### LINKED ISSUE #154: chore: extract DB migrations to a one-shot startup step before scaling app replicas
## Problem
The project's internal roadmap (deploy/scaling notes) explicitly flags a latent bug in the current deploy setup:

> One piece is worth doing standalone whenever convenient, because it's a latent bug: the Docker `CMD` runs `migrate && ensure-tables && server.js` in **every** container, so the day there's ever >1 replica the migrations race on startup. Fix = extract migrations to a one-shot step before the app starts (`Dockerfile` CMD + startup scripts). Low priority while single-container.

Right now the app runs as a single container (one dedicated server runs app + Redis + Postgres together), so this isn't biting today. But it's a real, understood, and already-diagnosed bug waiting to reappear the moment a second app replica is ever added, and it has no tracking issue yet.

## Proposed fix
- Extract the migration step (`bun scripts/migrate.mjs`) out of the per-container startup `CMD` in the `Dockerfile`.
- Run migrations as a separate one-shot step (e.g. a dedicated init container, a pre-deploy hook in Coolify, or a startup script that only the first/designated replica runs) before any app replica starts serving traffic.
- Confirm `ensure-tables` (if it's a separate idempotent step from `migrate`) is safe to still run per-container, or fold it into the same one-shot step if it's also not safe under concurrent execution.

## Why this matters
This is explicitly pre-diagnosed as a race condition, not a hypothetical — multiple containers would run `drizzle-orm/postgres-js/migrator` concurrently against the same database on startup. It's low priority while the app is single-container (per the roadmap's current scaling stance — scale-out is currently parked), but worth fixing opportunistically before it's ever needed under time pressure during an actual scale-out.

nazarli-shabnam:
## Summary
Closes #154.

The Docker `CMD` ran `bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs && bun server.js` in every container. This is safe today because the app runs as a single container/replica, but it's a pre-diagnosed race condition waiting to reappear the moment a second app replica is ever started: multiple containers would run `drizzle-orm/postgres-js/migrator` concurrently against the same database on boot.

## Changes
- **Dockerfile**: `CMD` now only starts the server (`bun server.js`). Migrations are no longer run per-container.
- **scripts/deploy.sh**: migrations now run as an explicit one-shot step — `docker compose run --rm app sh -c "bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs"` — after building the new image and ensuring `db`/`redis` are up, but before the `app` container is started. `docker compose run` respects the existing `depends_on: condition: service_healthy` for `db`/`redis`, so no extra wait-loop was needed.
- This also makes migration failures more visible: they now fail the deploy script directly (`set -euo pipefail`) with a scoped log message, instead of being buried inside the app container's boot sequence.

## Why this is safe
- `migrate.mjs` and `ensure-tables.mjs` are both idempotent (Drizzle's migration tracking table + `CREATE ... IF NOT EXISTS` DDL respectively), so running them once per deploy, before the app starts, is equivalent to the old behavior for the current single-replica setup — just no longer racy if replicas are ever added.
- `docker-compose.yml` is unchanged; no schema/runtime behavior changes, no new dependencies.

## Test plan
- [x] `bun run typecheck` (no source changes, included for completeness)
- [x] Traced `deploy.sh` step-by-step against `docker-compose.yml`'s `depends_on`/healthcheck config to confirm `docker compose run --rm app` waits for `db`/`redis` health the same way `up -d --build` did
- [ ] Infra-only change (Dockerfile + deploy shell script) with no existing automated coverage for deploy.sh; recommend a dry run against a staging-like compose stack before the next production deploy

## Summary by CodeRabbit

* **Deployment**
  * Updated container startup to run only the application server, without performing database migrations or table checks on boot.
  * Adjusted the deployment flow to run migrations and table-ensuring as a one-shot step before starting the application container.
  * Lowered the chance of migration conflicts across multiple running instances.
  * Kept existing health checks and rollback behavior for safer releases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,8 @@ COPY --from=deps /app/node_modules/@google/generative-ai ./node_modules/@google/
 
 USER nextjs
 EXPOSE 3000
-# Run pending migrations (idempotent) then start the server.
-# This ensures new tables are always created on deploy without a manual step.
-CMD ["sh", "-c", "bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs && bun server.js"]
+# Migrations run as a one-shot step in scripts/deploy.sh before this container
+# starts, not here — running migrate.mjs from every replica's CMD would race
+# multiple `drizzle-orm/postgres-js/migrator` runs against the same DB once
+# there's ever more than one app replica.
+CMD ["bun", "server.js"]
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -21,7 +21,10 @@ echo "==> Ensuring database and redis are running..."
 $COMPOSE up -d db redis
 
 echo "==> Running database migrations (one-shot, before the app starts)..."
-$COMPOSE run --rm app sh -c "bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs"
+# --name avoids colliding with the still-running `open-hikmah-app` container
+# (fixed container_name in docker-compose.yml) — the old app container isn't
+# stopped until `up -d app` below replaces it.
+$COMPOSE run --rm --name open-hikmah-migrate app sh -c "bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs"
 
 echo "==> Starting app container..."
 $COMPOSE up -d app
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -14,8 +14,17 @@ git pull origin main
 echo "==> Tagging current image as :previous (for rollback)..."
 docker tag open-hikmah-app:latest open-hikmah-app:previous 2>/dev/null || true
 
-echo "==> Building new image and starting containers..."
-$COMPOSE up -d --build
+echo "==> Building new app image..."
+$COMPOSE build app
+
+echo "==> Ensuring database and redis are running..."
+$COMPOSE up -d db redis
+
+echo "==> Running database migrations (one-shot, before the app starts)..."
+$COMPOSE run --rm app sh -c "bun scripts/migrate.mjs && bun scripts/ensure-tables.mjs"
+
+echo "==> Starting app container..."
+$COMPOSE up -d app
 
 echo "==> Waiting for health check (up to 60s)..."
 HEALTHY=false


nazarli-shabnam:
### Code review

Found 1 issue (fixed in a follow-up commit before this comment):

1. `docker compose run --rm app ...` for the one-shot migration step collided with the service's fixed `container_name: open-hikmah-app` — since the currently-running app container isn't stopped until the later `up -d app` step, every redeploy after the first would hit a "container name already in use" error and the migration step (now a hard prerequisite) would block the whole deploy.

Fixed in [e1f51a0](https://github.com/OpenHikmah/openhikmah-web/blob/e1f51a0a4760a9e699a24d447be085e99987fb43/scripts/deploy.sh#L20-L25) by giving the one-shot container its own name (`--name open-hikmah-migrate`).

Also flagging (pre-existing, not introduced by this PR, out of scope to fix here): `docker-compose.yml`'s redis healthcheck (`redis-cli ping`) doesn't pass `-a "$REDIS_PASSWORD"`, so if `REDIS_PASSWORD` is ever set, redis never reports healthy — which would make `depends_on: condition: service_healthy` block indefinitely for both the app service and this PR's new migration step. Worth a separate follow-up issue.

nazarli-shabnam:
Thanks — this is a fair point, but it's a larger scope than issue #154 asked for, so I don't think it should block this PR:

Issue #154 was specifically about the migration race condition when there's more than one app replica. The app runs as a single container/replica today (per the issue's own framing: "low priority while single-container"), so there's no live traffic being served by a second replica during migration right now — traffic-quiescing only matters once actual horizontal scale-out happens, which the project's current hosting setup doesn't do.

Coordinating a traffic-quiesce window and building a tested schema-rollback path is a meaningfully bigger effort (load balancer / proxy coordination, migration reversibility tooling, a rollback drill) than "move migrate.mjs out of the per-container CMD," and conflates two different problems: the migration race (fixed here) vs. zero-downtime deploys generally (not yet a requirement, since it's single-replica).

I'd suggest filing that as its own follow-up issue for whenever scale-out is actually planned, rather than expanding this PR's scope. Happy to be corrected if there's a plan to add a second replica sooner than I'm assuming.

PR #304, opened 2026-07-26
fix: enforce csp-report body size cap while streaming, not after buffering

### LINKED ISSUE #254: fix: /api/csp-report buffers full request body before enforcing the size cap
## Problem
The CSP violation report endpoint checks \`Content-Length\` before reading the body, then re-checks length after \`await req.text()\` has already fully buffered the body in memory:

\`\`\`ts
const contentLength = req.headers.get("content-length");
if (contentLength && Number(contentLength) > MAX_BODY_BYTES) {
  return new NextResponse(null, { status: 413 });
}

const text = await req.text(); // <-- entire body already read into memory here
if (text.length > MAX_BODY_BYTES) {
  return new NextResponse(null, { status: 413 });
}
\`\`\`

\`Content-Length\` is attacker-controlled and easy to bypass (omit the header, use chunked transfer-encoding). The second check happens only *after* the full body has been materialized as a string.

## Impact
This is a public, unauthenticated endpoint (browsers POST to it directly per the CSP \`report-to\`/\`report-uri\` directive), and per the code comment it's explicitly attacker-influenced ("a page under CSP attack controls what gets reported"). Without a true streaming cap, a malicious or misbehaving client can send an arbitrarily large body and force the server to buffer it fully before rejecting it — a cheap memory-pressure DoS vector.

## Location
\`app/api/csp-report/route.ts:24-30\`

## Suggested fix
Read the request body as a stream and abort once bytes exceed \`MAX_BODY_BYTES\`, instead of trusting \`Content-Length\` as the sole gate before calling \`req.text()\`. Alternatively, wrap the stream read with a byte-counting reader that throws/returns 413 as soon as the cap is exceeded, without buffering past it.

nazarli-shabnam:
## Summary
Closes #254.

`POST /api/csp-report` checked `Content-Length` before parsing, then re-checked size only after `await req.text()` had already fully buffered the body in memory. `Content-Length` is attacker-controlled (easy to omit or under-report via chunked transfer-encoding), so the real cap only ever applied post-buffering — a hostile or misbehaving client could force the server to materialize an arbitrarily large body before rejecting it. This is a public, unauthenticated endpoint that the code's own comment already notes is attacker-influenced.

## Changes
- `app/api/csp-report/route.ts`: replaced the `Content-Length` pre-check + `req.text()` with a streaming reader (`readCappedBody`) that reads the body chunk by chunk via `req.body.getReader()`, tracks total bytes, and cancels the stream (`reader.cancel()`) the moment the running total exceeds the 32KB cap — never materializing more than the cap's worth of the body.
- Added `__tests__/api/csp-report.test.ts` covering: valid legacy/reports+json payloads, unparseable JSON (204, not an error), an oversized body (413), an oversized body with a spoofed/under-reported `Content-Length` (413 — the actual attack this fixes), and a streaming test asserting the reader is cancelled well before pulling the full oversized body (proves no full-buffering happens).

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx vitest run __tests__/api/csp-report.test.ts` (6/6 passing)

### DIFF
--- a/__tests__/api/csp-report.test.ts
+++ b/__tests__/api/csp-report.test.ts
@@ -1,0 +1,92 @@
+import { describe, it, expect } from "vitest";
+import { NextRequest } from "next/server";
+import { POST } from "@/app/api/csp-report/route";
+
+const MAX_BODY_BYTES = 32 * 1024;
+
+function reqWithBody(body: string) {
+  return new NextRequest("http://localhost/api/csp-report", {
+    method: "POST",
+    headers: { "Content-Type": "application/csp-report" },
+    body,
+  });
+}
+
+/** A request whose declared Content-Length lies about the true body size. */
+function reqWithSpoofedLength(body: string, declaredLength: number) {
+  return new NextRequest("http://localhost/api/csp-report", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/csp-report",
+      "Content-Length": String(declaredLength),
+    },
+    body,
+  });
+}
+
+describe("POST /api/csp-report", () => {
+  it("accepts a well-formed legacy csp-report", async () => {
+    const res = await POST(
+      reqWithBody(JSON.stringify({ "csp-report": { "violated-directive": "script-src" } }))
+    );
+    expect(res.status).toBe(204);
+  });
+
+  it("accepts a well-formed reports+json batch", async () => {
+    const res = await POST(
+      reqWithBody(JSON.stringify([{ type: "csp-violation", body: { blockedURL: "x" } }]))
+    );
+    expect(res.status).toBe(204);
+  });
+
+  it("returns 204 (not an error) for unparseable JSON", async () => {
+    const res = await POST(reqWithBody("not json"));
+    expect(res.status).toBe(204);
+  });
+
+  it("treats a request with no body as an empty (unparseable) payload", async () => {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (fixed in a follow-up commit before this comment):

1. The new `readCappedBody` had an untested branch (`req.body?.getReader()` returning undefined for a bodyless request) — new logic should ship with test coverage for all its branches per [AGENTS.md](https://github.com/OpenHikmah/openhikmah-web/blob/4420b006d7fa762b8f411c120637bd1fe6f8dea7/AGENTS.md#L46)'s testing bar ("new logic requires new tests in the same PR").

Fixed in [4420b00](https://github.com/OpenHikmah/openhikmah-web/blob/4420b006d7fa762b8f411c120637bd1fe6f8dea7/__tests__/api/csp-report.test.ts#L47-L51) by adding a test for a bodyless POST.

No other issues found: the streaming cap correctly avoids the multi-byte-UTF-8-truncation, off-by-one, and unhandled-rejection failure modes checked, and the removal of the `Content-Length` pre-check doesn't weaken the [defensive-cap comment](https://github.com/OpenHikmah/openhikmah-web/blob/4420b006d7fa762b8f411c120637bd1fe6f8dea7/app/api/csp-report/route.ts#L7-L9) it implements — it satisfies it more strictly than the code it replaces.

PR #305, opened 2026-07-26
chore: gate /api/metrics behind requireAdmin

### LINKED ISSUE #255: chore: /api/metrics exposes internal counters/latencies with no auth gate
## Problem
\`GET /api/metrics\` returns process uptime, Redis/DB reachability + latency, and in-process counters (including \`ratelimit_fail_open\`, \`ratelimit_block\`, AI call counts) to any caller, with no authentication:

\`\`\`ts
export async function GET() {
  const [redis, database] = await Promise.all([probeRedis(), probeDb()]);
  return NextResponse.json({
    uptimeSeconds: uptimeSeconds(),
    redis,
    db: database,
    counters: counterSnapshot(),
  });
}
\`\`\`

This is inconsistent with \`/api/admin/infra\`, which is correctly gated behind \`requireAdmin\` for what appears to be overlapping/related data.

## Impact
The code comment states the route "Exposes no secrets or user data," which is true, but it still leaks internal operational signals to any anonymous caller: DB/Redis latency (useful for timing/availability recon), rate-limit block/fail-open counts (reveals whether rate limiting is currently degraded), and AI call volume. This is low severity but is unnecessary information disclosure for a route with no access control at all.

## Location
\`app/api/metrics/route.ts:40\`

## Suggested fix
Either gate the route behind \`requireAdmin\` (matching \`/api/admin/infra\`), or if it's intentionally meant to be a public health-check/scrape target (e.g. for uptime monitoring), keep it public but strip the more sensitive counters (rate-limit and AI-call counts) and keep only basic up/down + latency, documenting the decision in a code comment.

nazarli-shabnam:
## Summary
Closes #255.

`GET /api/metrics` returned process uptime, Redis/DB reachability + latency, and in-process counters (including `ratelimit_fail_open`, `ratelimit_block`, AI call counts) to any anonymous caller. `/api/admin/infra` already gates materially overlapping data behind `requireAdmin`; this route had no access control at all.

## Changes
- `app/api/metrics/route.ts`: gated `GET` behind `requireAdmin`, matching the `/api/admin/infra` pattern.
- Added `__tests__/api/metrics.test.ts`: verifies non-admin/unauthenticated callers get the guard's own response, admin callers get the full snapshot, and the db-unreachable branch is exercised.

No public/anonymous consumer of this route was found in the codebase (checked for fetch/curl references beyond code comments), so this isn't expected to break an external health-check integration — `docker-compose.yml`'s healthcheck uses `/api/health`, a separate route.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx vitest run __tests__/api/metrics.test.ts` (3/3 passing)

### DIFF
--- a/__tests__/api/metrics.test.ts
+++ b/__tests__/api/metrics.test.ts
@@ -1,0 +1,76 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest, NextResponse } from "next/server";
+import type { User } from "@/lib/infra/db/schema";
+
+vi.mock("@/lib/admin/admin-auth", () => ({
+  requireAdmin: vi.fn(),
+}));
+
+const { mockExecute } = vi.hoisted(() => ({
+  mockExecute: vi.fn(() => Promise.resolve(undefined)),
+}));
+vi.mock("@/lib/infra/db", () => ({ db: { execute: mockExecute } }));
+
+const { mockRedisEnabled, mockGetRedis } = vi.hoisted(() => ({
+  mockRedisEnabled: vi.fn(() => false),
+  mockGetRedis: vi.fn<() => { ping: () => Promise<string> } | null>(() => null),
+}));
+vi.mock("@/lib/infra/redis", () => ({
+  redisEnabled: mockRedisEnabled,
+  getRedis: mockGetRedis,
+}));
+
+const { mockCounterSnapshot, mockUptimeSeconds } = vi.hoisted(() => ({
+  mockCounterSnapshot: vi.fn(() => ({ "csp.violation": 3 })),
+  mockUptimeSeconds: vi.fn(() => 123),
+}));
+vi.mock("@/lib/infra/metrics", () => ({
+  counterSnapshot: mockCounterSnapshot,
+  uptimeSeconds: mockUptimeSeconds,
+}));
+
+import { GET } from "@/app/api/metrics/route";
+import { requireAdmin } from "@/lib/admin/admin-auth";
+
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

No blocking issues found. `requireAdmin` is applied before the Redis/DB probes run (no pre-auth timing leak), its result is checked correctly, and the [JSDoc update](https://github.com/OpenHikmah/openhikmah-web/blob/da386b59dd7379c7c742c14bb7c2981e689d717e/app/api/metrics/route.ts#L35-L42) accurately describes the new behavior.

One thing worth surfacing since this touches auth on a previously-public route: PR #89 (the PR that originally introduced this endpoint) had a CodeRabbit comment flagging the missing access control and suggesting a lightweight `METRICS_TOKEN` bearer guard rather than full admin auth — for a route intended to also be scrapeable by an external monitoring tool (Prometheus-style) without a QF OAuth session. That comment went unaddressed and unanswered at the time.

I checked the codebase for any such external scraper/consumer and found none — no Prometheus config, no other fetch site beyond code comments — and issue #255 itself explicitly offers `requireAdmin` as one of two sanctioned options (the other being "keep public, strip sensitive counters"). Going with `requireAdmin` is consistent with that issue text and matches `/api/admin/infra`'s existing pattern. Flagging here only so it's visible in review: if there's an unauthenticated external monitoring integration this repo doesn't track in-tree, this change would silently break it, and a lighter `METRICS_TOKEN` guard would be the fix instead.

PR #307, opened 2026-07-26
fix: validate data shape per kind in admin PATCH /api/admin/names

### LINKED ISSUE #257: fix: admin PATCH /api/admin/names doesn't validate data shape per kind
## Problem
\`PATCH /api/admin/names\` persists admin-submitted \`data\` for a given \`(slug, kind)\` pair without validating that its shape matches what the public-facing route expects for that \`kind\` (e.g. \`reflection\` should be a string, \`verses\`/\`pairings\` should be arrays).

## Impact
A malformed admin edit (typo in the request body, wrong type, wrong field name) is accepted and persisted as-is, silently breaking the public name page's rendering for all visitors until someone notices and manually corrects the record. There's no immediate validation error to catch the mistake at edit time.

## Location
\`app/api/admin/names/route.ts:60\`

## Suggested fix
Validate \`data\`'s shape against the expected type for the given \`kind\` before writing (e.g. a small per-kind Zod schema or manual type guard), returning a 400 with a clear message if it doesn't match, instead of persisting unchecked.

nazarli-shabnam:
## Summary
Closes #257.

`PATCH /api/admin/names` persisted admin-submitted `data` for a given `(slug, kind)` without validating its shape matched what the public-facing routes expect for that `kind`. A malformed edit (typo, wrong type, wrong field name) would silently break the public name page's rendering for all visitors.

## Changes
`app/api/admin/names/route.ts`: added a per-kind type guard, matching the shapes each public route actually produces/consumes:
- `reflection` → plain string (`app/api/names/[slug]/reflection/route.ts`)
- `pairings` → array of `{ name, transliteration, arabic, explanation }` (`app/api/names/[slug]/pairings/route.ts`'s `Pairing` type)
- `verses` → array of `{ ref, surah, ayah, arabicText, translation, surahName, surahNameArabic, reason }` (`app/api/names/[slug]/verses/route.ts`'s `NameVerse` type)

`PATCH` now returns 400 with a clear message when `data` doesn't match the expected shape for `kind`, before writing. No new dependency — matches the existing manual-type-guard pattern already used in `lib/canvas/share-canvas.ts` (`isValidNode`) rather than introducing Zod for one route.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx vitest run __tests__/api/admin/names.test.ts` (16/16 passing) — added tests for valid/invalid data per kind (reflection non-string, verses non-array and wrong-field-type, pairings missing a required field), and updated the pre-existing "no cached row" / "updates the cached content" tests to use shape-valid fixtures

## Summary by CodeRabbit

* **New Features**
  * Added validation for admin content updates, ensuring reflections, pairings, and verses use supported data formats.
  * Invalid update requests now receive a clear 400 error without modifying stored content.

* **Bug Fixes**
  * Improved handling of cached verse updates and ensured valid content is returned correctly.
  * Strengthened test coverage for accepted and rejected update payloads.

### DIFF
--- a/__tests__/api/admin/names.test.ts
+++ b/__tests__/api/admin/names.test.ts
@@ -66,7 +66,7 @@ function del(query?: string) {
 beforeEach(() => {
   vi.mocked(requireAdmin).mockResolvedValue(admin);
   mockSelect.mockReturnValue(makeDbChain([]));
-  mockUpdate.mockReturnValue(makeDbChain([]));
+  mockUpdate.mockClear().mockReturnValue(makeDbChain([]));
   mockDelete.mockReturnValue(makeDbChain([]));
   vi.mocked(logAdminAction).mockClear();
 });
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, fixed in a follow-up commit:

1. The `pairings` validator only checked `typeof === "string"`, so `name: ""` (an unresolved slug reference) passed validation. PR #89's CodeRabbit review flagged this exact shape on the AI-generation path (\`app/api/names/[slug]/pairings/route.ts\`) — an empty \`name\` still got cached and served. The admin route would have let the same bad shape back in.

Fixed in [b1235c8](https://github.com/OpenHikmah/openhikmah-web/blob/b1235c8be3fdb236b4d3e5df8fbd6c1c5dda07ef/app/api/admin/names/route.ts#L12-L17): all three validators now require non-empty (trimmed) strings, not just the right type.

One minor, non-blocking note: two test fixtures use `arabicText: "arabic"` / `"a"` as placeholder Arabic text rather than a real string — [AGENTS.md](https://github.com/OpenHikmah/openhikmah-web/blob/b1235c8be3fdb236b4d3e5df8fbd6c1c5dda07ef/AGENTS.md#L51) asks for real/plausible Arabic even in test fixtures for that field. Low severity since it's shape-validation test data, not user-facing content — flagging for awareness rather than blocking.

No other issues found: the guards' field lists match `NameVerse`/`Pairing`'s real shapes exactly (verified field-by-field), the `VALIDATORS[kind as Kind]` cast is safe since `kind` is already narrowed by the earlier `KINDS.includes` check, and validation runs before the DB write.

nazarli-shabnam:
Fixed in 931a500 — `isValidVerses` now uses `Number.isFinite` instead of `typeof === "number"` for `surah`/`ayah`, with a regression test using a raw JSON body (`1e400`) since `JSON.stringify(Infinity)` can't reproduce the bug through the normal test helper. Thanks for catching this.

PR #309, opened 2026-07-26
fix: dedup redisSubscribe listener registration per channel

### LINKED ISSUE #259: fix: redisSubscribe re-registers a message listener per call with no dedup
## Problem
\`redisSubscribe(channel, onMessage)\` adds a brand-new \`"message"\` listener to the shared \`subscriberClient\` on every call, filtered internally by \`ch === channel\`. There's no tracking of already-registered channels/handlers and no removal:

\`\`\`ts
subscriberClient.on("message", (ch, msg) => {
  if (ch === channel) onMessage(msg);
});
\`\`\`

## Impact
Currently masked because the sole caller (\`lib/auth/social-auth.ts\`, lines ~96-97) guards itself with its own \`subscribed\` boolean, so \`redisSubscribe\` only ever gets called once per channel in practice. But the primitive itself is unsafe by default: any future caller that invokes \`redisSubscribe\` more than once for the same channel (e.g. a hot-reload path, a second feature wiring the same invalidation channel) will stack listeners indefinitely, each stacked listener re-invoking \`onMessage\` redundantly per message, and eventually tripping Node's default max-listeners warning (10).

## Location
\`lib/infra/redis.ts:122-144\`

## Suggested fix
Track registered channels/handlers inside \`redisSubscribe\` itself (e.g. a \`Map<channel, Set<handler>>\`), and register the underlying \`"message"\` listener on \`subscriberClient\` at most once per subscriber client, dispatching to all registered handlers for a channel from that single listener.

nazarli-shabnam:
## Summary
Closes #259.

`redisSubscribe(channel, onMessage)` added a brand-new `"message"` listener to the shared `subscriberClient` on every call, filtered internally by `ch === channel`, with no tracking of already-registered channels/handlers. Currently masked because the sole caller (`lib/auth/social-auth.ts`) guards itself with its own `subscribed` boolean — but the primitive itself is unsafe by default: any future caller invoking `redisSubscribe` more than once for the same channel would stack listeners indefinitely, each stacked listener redundantly re-invoking `onMessage` per message, eventually tripping Node's max-listeners warning.

## Changes
`lib/infra/redis.ts`:
- Handlers are now tracked per channel in a `Map<string, Set<handler>>`, persisted on `globalThis` (same pattern as `subscriberClient` itself) so a dev hot-reload doesn't orphan handlers registered before the reload.
- The underlying Redis channel is only `subscribe()`'d once per distinct channel (repeat calls for the same channel just add another handler to the existing set).
- The single `"message"` dispatcher listener is registered at most once per subscriber client (guarded by `listenerCount("message") === 0`), and dispatches to every handler registered for the incoming message's channel.

No signature change — `redisSubscribe`'s call sites in `lib/auth/social-auth.ts` are untouched.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx vitest run __tests__/lib/infra/redis.test.ts` (8/8 passing) — added tests asserting: only one `"message"` listener is ever registered across repeated calls, the underlying channel is `subscribe()`'d only once even when called repeatedly for it, and an incoming message dispatches to every handler registered for that channel but not to handlers on other channels
- [x] `bunx vitest run __tests__/lib/auth` (31/31 passing) — confirms the sole real caller (`social-auth.ts`) is unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redis pub/sub message handling across hot reloads.
  * Prevented duplicate channel subscriptions and message listeners.
  * Ensured all registered handlers receive messages for their subscribed channel without affecting other channels.
  * Strengthened subscription behavior and test isolation.

### DIFF
--- a/__tests__/lib/infra/redis.test.ts
+++ b/__tests__/lib/infra/redis.test.ts
@@ -9,15 +9,31 @@ const behavior = vi.hoisted(() => ({
   del: vi.fn(),
   ping: vi.fn(),
   multiExec: vi.fn(),
+  subscribe: vi.fn((..._args: unknown[]) => Promise.resolve()),
 }));
 
 vi.mock("ioredis", () => {
+  // Minimal inline emitter (avoids importing "node:events" inside a hoisted
+  // vi.mock factory) — just enough for redisSubscribe's on()/emit()/listenerCount().
   class MockRedis {
+    private listeners = new Map<string, Set<(...args: unknown[]) => void>>();
     constructor(...args: unknown[]) {
       behavior.ctor(...args);
     }
-    on() {
+    on(event: string, handler: (...args: unknown[]) => void) {
+      let set = this.listeners.get(event);
+      if (!set) {
+        set = new Set();
+        this.listeners.set(event, set);
+      }
+      set.add(handler);
       return this;
+    }
+    emit(event: string, ...args: unknown[]) {
+      this.listeners.get(event)?.forEach((handler) => handler(...args));
+    }
+    listenerCount(event: string) {
+      return this.listeners.get(event)?.size ?? 0;
     }
     get(...a: unknown[]) {
       return behavior.get(...a);
@@ -39,6 +55,12 @@ vi.mock("ioredis", () => {
       };
       return chain;
     }
+    subscribe(...a: unknown[]) {
+      return behavior.subscribe(...a);
+    }
+    duplicate() {
+      return new MockRedis();
+    }
   }
   return { default: MockRedis };
 });
@@ -48,6 +70,8 @@ vi.mock("ioredis", () => {
 beforeEach(() => {
   vi.resetModules();
   delete (globalThis as { __redis?: unknown }).__redis;
+  delete (globalThis as { __redisSub?: unknown }).__redisSub;
+  delete (globalThis as { __redisSubHandlers?: unknown }).__redisSubHandlers;
   delete process.env.REDIS_URL;
   behavior.ctor.mockReset();
   behavior.get.mockReset();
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review the PR!

nazarli-shabnam:
### Code review

Found 1 issue, fixed in a follow-up commit:

1. Deduping the `SUBSCRIBE` call itself (only issuing it the first time a channel was seen, gated on the `Map` entry) removed the self-healing retry the old code had — if the first `SUBSCRIBE` attempt failed (e.g. Redis unreachable at startup), the channel's entry would already exist in the `Map`, so no later call would ever retry it, permanently stranding that channel's handlers for the life of the process. The old code re-issued `SUBSCRIBE` on every call, so a later invocation would naturally retry.

Fixed in [edf93f3](https://github.com/OpenHikmah/openhikmah-web/blob/edf93f383ea24a6c17c1e69dc670bfd2a8f9ce6a/lib/infra/redis.ts#L155-L166): `SUBSCRIBE` is issued on every `redisSubscribe` call again (matching prior behavior — Redis/ioredis treat a repeat `SUBSCRIBE` to an already-subscribed channel as a no-op), while the actual reported bug — stacked `"message"` listeners — stays fixed via the `listenerCount("message") === 0` guard, which is unaffected by this change.

No other issues found: the dispatcher closure re-resolves `channelHandlers()` fresh on every message rather than capturing a stale reference (correct across dev hot-reloads, since both the client and the handler map persist via `globalThis`), and there's no race between the map check and the subscribe call since `redisSubscribe` runs fully synchronously.

PR #310, opened 2026-07-26
fix: draw edge instead of silently dropping expansion to an existing node

### LINKED ISSUE #260: fix: canvas expansion silently skips connections to verses already on canvas, no edge drawn
## Problem
When \`runExpansion\` walks the AI-returned connections for a node, any connection whose target verse ref is already present as a node anywhere on the canvas is simply skipped:

\`\`\`ts
if (hasNode(conn.ref)) continue;
\`\`\`

No edge is drawn from the source node to the pre-existing duplicate node, and nothing is surfaced to the user explaining why fewer nodes appeared than expected.

## Impact
A user who expands a node and gets, say, 4 AI-returned results where 2 already exist elsewhere on the canvas will see only 2 new nodes appear, with no edge linking the source to the 2 pre-existing verses, and no message explaining the discrepancy. Because \`connections.length > 0\` in this case, the existing \`ExhaustedExpansionError\` path never fires either, so the loading spinner just completes with an outcome that looks incomplete/broken but isn't actually an error.

## Location
\`components/canvas/HikmahCanvas.tsx:118-162\` (specifically the \`continue\` at line ~123)

## Suggested fix
When a connection's target already exists as a node, still draw an edge from the source node to the existing node (via \`getNodeByRef\`/\`addConnectionEdge\`) instead of dropping the connection entirely. This preserves the semantic link the AI identified without creating a duplicate node.

nazarli-shabnam:
## Summary
Closes #260.

When `runExpansion` walked AI-returned connections for a node, any connection whose target verse ref was already present as a node anywhere on the canvas was simply skipped (`continue`) — no edge drawn, no message explaining the discrepancy. Because `connections.length > 0` in this case, the `ExhaustedExpansionError` path never fired either, so the outcome looked incomplete/broken without being flagged as an error.

## Changes
- `lib/canvas/canvas-layout.ts`: added `buildExistingNodeEdge(sourceNodeId, existingNodeId, conn)`, a small pure helper (alongside the existing `findFreeSlot`/`viewportCenter` extractions from this same component) that builds the edge to draw when a connection's target already exists as a node — returning `null` for a self-loop (the AI returning the source's own ref).
- `components/canvas/HikmahCanvas.tsx`: when `hasNode(conn.ref)` is true, looks up the existing node via `getNodeByRef` and draws the edge via `addConnectionEdge` instead of dropping the connection. `addConnectionEdge` already dedupes repeat source/target pairs (either direction), so re-expanding the same node is a safe no-op — no new duplicate-edge logic was needed there.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx vitest run __tests__/lib/canvas/canvas-layout.test.ts __tests__/store/canvas.test.ts` (42/42 passing) — added tests for `buildExistingNodeEdge`: builds the correct edge, truncates the label to 60 chars while keeping the full reason, and returns `null` for a self-loop

## Summary by CodeRabbit

* **New Features**
  * Canvas expansion now preserves suggested connections to nodes that already exist, avoiding duplicate nodes.
  * Existing-node connections use the correct connection type and show a shortened label while keeping the full explanation available.
  * Self-referential connections are excluded to prevent invalid canvas links.
* **Tests**
  * Added unit coverage for existing-node edge construction, label truncation behavior, connection field correctness, and self-loop handling.

### DIFF
--- a/__tests__/lib/canvas/canvas-layout.test.ts
+++ b/__tests__/lib/canvas/canvas-layout.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect } from "vitest";
 import {
   findFreeSlot,
   viewportCenter,
-  buildExistingNodeEdge,
+  buildConnectionEdge,
   NODE_WIDTH,
   NODE_HEIGHT,
   NODE_GAP,
@@ -86,11 +86,11 @@ describe("viewportCenter", () => {
   });
 });
 
-describe("buildExistingNodeEdge", () => {
+describe("buildConnectionEdge", () => {
   const conn = { kind: "thematic" as const, reason: "Both describe divine mercy." };
 
   it("builds an edge from the source node to the existing node", () => {
-    const edge = buildExistingNodeEdge("node-a", "node-b", conn);
+    const edge = buildConnectionEdge("node-a", "node-b", conn);
     expect(edge).toEqual({
       id: "edge-node-a-node-b",
       source: "node-a",
@@ -102,12 +102,12 @@ describe("buildExistingNodeEdge", () => {
 
   it("truncates the label to 60 characters but keeps the full reason", () => {
     const longReason = "x".repeat(100);
-    const edge = buildExistingNodeEdge("node-a", "node-b", { kind: "root", reason: longReason });
+    const edge = buildConnectionEdge("node-a", "node-b", { kind: "root", reason: longReason });
     expect(edge!.data.label).toHaveLength(60);
     expect(edge!.data.reason).toBe(longReason);
   });
 
   it("returns null for a self-loop (existing node is the source itself)", () => {
-    expect(buildExistingNodeEdge("node-a", "node-a", conn)).toBeNull();
+    expect(buildConnectionEdge("node-a", "node-a", conn)).toBeNull();
   });
 });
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
### Code review

No blocking issues. Traced the whole call path (getNodeByRef/hasNode synchronicity, the staggered-reveal timing, the self-loop guard against the AI/connection-generation pipeline's own \`c.ref !== fromRef\` filters, and \`addConnectionEdge\`'s existing dedup) — all correct and safe.

Two minor nitpicks, both fixed in a follow-up commit rather than left as-is:

1. The new-node edge (\`addConnectionEdge({...})\` for the freshly-created node) built the same edge shape inline, duplicating \`buildExistingNodeEdge\`'s logic — a drift risk between the two paths.
2. \`lib/canvas/canvas-layout.ts\`'s file docstring still said "collision-aware node placement," not covering the new edge-construction responsibility.

Fixed in [700af1b](https://github.com/OpenHikmah/openhikmah-web/blob/700af1b78494c503d22464dbe6b1e7ec3477aa1d/lib/canvas/canvas-layout.ts#L113-L136): renamed to \`buildConnectionEdge\` and used it for both the new-node and existing-node edges in \`runExpansion\`, and updated the docstring.

nazarli-shabnam:
@coderabbitai review the PR!

nazarli-shabnam:
@coderabbitai review the PR!

nazarli-shabnam:
@coderabbitai review the PR!

PR #316, opened 2026-07-28
feat: add stadium wave animation to landing page

Mubashir78:
Add a canvas-based stadium wave effect to the landing page that travels diagonally (top-left to bottom-right) as a looping animation band.

https://github.com/user-attachments/assets/349987fd-6a87-41cf-8214-e25b4c48c511


nazarli-shabnam:
@Mubashir78 what about after sign in? the page stays the same as this or prev?

Mubashir78:
> @Mubashir78 what about after sign in? the page stays the same as this or prev?

It changes to "Welcome back" and shows my canvas list. The wave animation will still be there, but not on the canvas page.

Mubashir78:
Basically the animation is always on the **home page** only, regardless of login or logout

Mubashir78:
The tunable values are at the top of `components/home/WaveBackground.tsx`:

```
const SIZE = 4;        // cube side length in px
const GAP = 5;         // gap between cubes in px
const SPEED = 0.08;    // base wave scroll speed
const BAND_WIDTH = 0.12;  // width of the diagonal band (normalized)
const MAX_HEIGHT = 5;  // max vertical cube lift in px
const PEAK_ALPHA = 0.15;  // peak opacity at the wave center
```

nazarli-shabnam (approved):
well done buds

PR #404, opened 2026-08-09
fix(build): skip TypeScript checking in next build to stop OOM deploy failures

nazarli-shabnam:
## Summary

- The last two production deploys (self-hosted Coolify/Docker, `openhikmah-server`) failed. The build log shows `next build` compiling successfully (~53s), then crashing silently — zero output — immediately after `Running TypeScript ...`, exit code 255. That's the exact signature of the build container being OOM-killed mid type-check, not a real compile error.
- Confirmed this isn't a real type error: CI's `lint-typecheck-test` job runs the full `bun run typecheck` on every push/PR and was green on the exact commit (`49c6bb8`) that failed to deploy.
- Next.js's own docs (`node_modules/next/dist/docs/01-app/02-guides/memory-usage.md`, "Disable static analysis" section) describe this precise symptom and recommend `typescript.ignoreBuildErrors: true` when a dedicated CI runner already handles the check — which this repo has.
- Added `typescript: { ignoreBuildErrors: true }` to `next.config.ts` with a comment explaining why, so the Docker build no longer redoes the memory-heavy type-check that CI already gates on GitHub's runners.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run typecheck` passes locally (confirms there are no real type errors being hidden by this change)
- [x] `bun run format:check` passes locally (pre-existing unrelated formatting warnings in `.superpowers/sdd/*` are untouched by this diff)
- [x] `bun run build` succeeds locally with the new config
- [ ] New tests added — N/A, this is a build-config change with no runtime code path

## Checklist

- [x] Conventional commits
- [ ] `.env.example` updated — N/A, no new environment variables
- [ ] `README.md` updated — N/A, no public interface change

## Risk / rollback note

`ignoreBuildErrors: true` means a future real type error would no longer block the *Docker production build* — but it will still block *merging to main*, since CI's `lint-typecheck-test` job (`bun run typecheck`) is unaffected and still required. If that CI gate is ever weakened or bypassed, this setting would let a type error reach production; this is an accepted tradeoff to stop deploys crashing today, not a first tolerance of type errors in general.

## Summary by CodeRabbit

* **Build Improvements**
  * Production builds can now complete even when TypeScript errors are present.
  * Type checking remains handled separately through the project’s CI workflow.

### DIFF
--- a/__tests__/next-config-build-policy.test.ts
+++ b/__tests__/next-config-build-policy.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from "vitest";
 import { readFileSync } from "fs";
 import path from "path";
+import nextConfig from "../next.config";
 
 // Regression test for the Docker-build OOM fix: next.config.ts skips
 // TypeScript checking during `next build` only because CI's separate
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

No issues found from the multi-agent review (CLAUDE.md/AGENTS.md compliance, bug scan, git history, prior PR precedent, and code-comment consistency all checked — nothing cleared the confidence bar for a real, actionable issue).

Addressed CodeRabbit's regression-test suggestion above regardless: added `__tests__/next-config-build-policy.test.ts` (7679aa1), which pins both halves of the tradeoff — `next.config.ts`'s `typescript.ignoreBuildErrors: true` and `.github/workflows/ci.yml`'s `bun run typecheck` step — so they can't silently drift apart in a future edit.

nazarli-shabnam:
Fixed in 981cb62: the regression test now checks the actual resolved \`next.config.ts\` export (\`nextConfig.typescript?.ignoreBuildErrors\`) instead of regexing the source text, and the CI-side check is scoped to the \`lint-typecheck-test\` job block specifically with an explicit non-comment check — so a commented-out block, a dead object, or a typecheck step moved to an unrelated job would all now fail the test. Verified by mutation-testing both failure modes locally before pushing.

PR #405, opened 2026-08-09
fix(stories): "Open on canvas" button can get stuck permanently disabled

### LINKED ISSUE #389: fix(stories): "Open on canvas" button can get stuck permanently disabled
## Summary

If navigation to the canvas page is interrupted after clicking "Open on canvas" from a story page, the button remains disabled indefinitely with no way to retry without a full reload.

## Evidence

`app/stories/[slug]/OpenOnCanvasButton.tsx:22-30`. `setPending(true)` (line 23) is never reset to `false` — it relies entirely on `router.push("/canvas")` (line 29) causing the component to unmount. If navigation is interrupted (blocked by the browser, or a client-side navigation error), the button stays disabled with no way to retry.

## Impact

Low-frequency but a real dead-end: a user whose navigation is interrupted must fully reload the story page to try again.

## Suggested fix

Reset `pending` to `false` in a `catch`/`finally` if `router.push` fails, or add a timeout-based fallback reset.

## Acceptance criteria

- An interrupted navigation does not leave the "Open on canvas" button permanently disabled.

nazarli-shabnam:
## Summary

Fixes #389.

- `app/stories/[slug]/OpenOnCanvasButton.tsx`: `setPending(true)` was set on click and only ever reset by the component unmounting when `router.push("/canvas")` succeeded. If navigation is interrupted (blocked, cancelled, or otherwise doesn't unmount the component), `pending` stayed stuck `true` forever, permanently disabling the button with no way to retry short of a full page reload.
- Wrapped the store-write + navigation in a `try`/`catch`: a synchronous throw now resets `pending` immediately.
- Added a 4-second fallback timeout that resets `pending` if the component is still mounted by then — long enough to never fire during a normal navigation (which unmounts the component well before that), but short enough that an interrupted navigation doesn't leave the user stuck.
- Guarded both resets with a `mountedRef` (same pattern already used in `components/canvas/HikmahCanvas.tsx`) so nothing sets state after unmount.

## Type of change

- [x] Bug fix

## AI / Theological changes

## Testing

- [x] New tests added: a fake-timers test confirming the button re-enables after the grace period if navigation never unmounts it, and a test confirming it re-enables immediately if adding a node throws synchronously.

## Checklist

- [x] Conventional commits
- [ ] `.env.example` updated — N/A
- [ ] `README.md` updated — N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * The “Open on canvas” button now reliably re-enables if navigation is interrupted or an error occurs.
  * Added a fallback that restores the button after four seconds when the page remains open.
  * Improved cleanup to prevent the button from becoming stuck in a loading state.
  * Failed actions are handled safely without triggering unintended navigation, allowing the action to be retried.

### DIFF
--- a/__tests__/components/stories/OpenOnCanvasButton.test.tsx
+++ b/__tests__/components/stories/OpenOnCanvasButton.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, act } from "@testing-library/react";
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import type { Verse } from "@/types/quran";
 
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue:

1. Silent catch block discarding the caught error with no logging (AGENTS.md says "no silent \`catch\` blocks... A failure should surface loudly") — every other catch block in the codebase logs via \`console.error\` before falling back; this was the only exception.

https://github.com/OpenHikmah/openhikmah-web/blob/9c1c17d0c297eb46390cc60b6ba55620c8e4c17/app/stories/%5Bslug%5D/OpenOnCanvasButton.tsx#L40-L44

Fixed in d2983d3.

A second candidate (the 4s fallback timeout could theoretically re-enable the button while a legitimately slow, not-failed, navigation is still in flight) was assessed but scored below the confidence bar for this report — real on paper, but low practical impact given \`/canvas\`'s typical load time and that a duplicate \`router.push\` just restarts the same navigation rather than corrupting anything. Not fixed in this PR; flagging for visibility only.

nazarli-shabnam:
On the silent-catch comment: fixed in d2983d3 — added \`console.error\` logging.

On the rethrow suggestion specifically: I tried it first, but verified with a test that it doesn't do what the suggestion intends. React error boundaries don't catch errors thrown from event handlers (only render/lifecycle errors), so rethrowing from this \`onClick\` handler doesn't route to \`app/stories/[slug]/error.tsx\` — it just becomes an unhandled exception (confirmed: the test failed with an "Uncaught Exception" vitest couldn't attribute to the synchronous \`fireEvent.click\` call). Logging via \`console.error\` is what actually satisfies "surface loudly" here without introducing a worse failure mode. Left a comment explaining this in the code.

On the second suggestion (test coverage for \`router.push\` throwing): added in d2983d3.

PR #406, opened 2026-08-09
fix(canvas): AI expansion silently drops a connection if any edge already exists between two nodes

### LINKED ISSUE #390: fix(canvas): AI expansion silently drops a connection if any edge already exists between two nodes
## Summary

When two verses are already connected on the canvas, a newly-discovered AI connection of a *different* kind between the same pair is silently dropped instead of being added or surfaced to the user.

## Evidence

`store/canvas.ts:217-238` (`addConnectionEdge`) no-ops if *any* edge already exists between the two node ids, regardless of `kind` (lines 219-224). In `components/canvas/HikmahCanvas.tsx`'s expansion loop (lines 154-164), when the target verse is already on the canvas, the code calls `addConnectionEdge` but never checks/uses its result.

## Impact

Reproduction: if a thematic edge already links two nodes and a later "by root" expansion identifies the same pair, the new connection is dropped with zero feedback — no new edge appears, and no notice is shown, even though the expansion loop otherwise reports a state for other failure modes (e.g. "no more connections").

## Suggested fix

Either allow multiple edges of different `kind` between the same node pair, or surface a visible (even subtle) indication when a connection was found but not added because one already exists.

## Acceptance criteria

- A user can tell, from the UI, when an AI-discovered connection was skipped because an edge already existed, rather than it silently vanishing.

nazarli-shabnam:
## Summary

Fixes #390.

- \`store/canvas.ts\` \`addConnectionEdge\` deduped edges purely by node pair (either direction), ignoring \`kind\`. If a thematic edge already linked two verses, a later "by root" (or any other kind) expansion that identified a connection between the same pair was silently dropped — no new edge, no feedback, even though the expansion loop already surfaces other failure states (e.g. "no more connections").
- \`addConnectionEdge\` now returns \`"added" | "duplicate-same-kind" | "duplicate-different-kind"\` instead of \`void\`, so callers can tell what happened.
- \`components/canvas/HikmahCanvas.tsx\`'s expansion loop now checks that result and, on \`"duplicate-different-kind"\`, surfaces a notice via the existing \`expansionNotice\`/\`noticeTimeoutRef\` transient-toast mechanism (already used for "no more connections" / "expansion failed") — extracted into a small \`showExpansionNotice\` helper shared by all three notice call sites to avoid tripling the timeout-scheduling logic.
- Added the new \`canvas.connectionExistsDifferentKind\` message key to all four locale files (en/tr/az/ru), with real translations (not English placeholders) for tr/az/ru, matching the existing terminology used elsewhere in the \`canvas\` namespace.
- Behavior is otherwise unchanged: still exactly one edge per node pair (the issue's suggested fix offered either "allow multiple edges" or "surface an indication" — went with surfacing, per the acceptance criteria, since allowing multiple simultaneous edges between the same pair is a larger UI/data-model change not asked for here).

## Type of change

- [x] Bug fix

## AI / Theological changes

## Testing

- [x] \`bun run test:ci\` passes locally
- [x] \`bun run typecheck\` passes locally
- [x] \`bun run lint\` passes locally
- [x] \`bun run format:check\` passes locally
- [x] New tests added:
  - Store: \`addConnectionEdge\` returns \`"added"\`/\`"duplicate-same-kind"\`/\`"duplicate-different-kind"\` correctly, including the reversed-pair case.
  - Component: a full expansion flow where the AI returns a different-kind connection to a verse already linked by an existing edge — asserts the notice text renders and no second edge is added.

## Checklist

- [x] Branch is up to date with \`main\`
- [x] Conventional commits
- [ ] \`.env.example\` updated — N/A
- [ ] \`README.md\` updated — N/A

### DIFF
--- a/__tests__/components/canvas/HikmahCanvas.timers.test.tsx
+++ b/__tests__/components/canvas/HikmahCanvas.timers.test.tsx
@@ -142,4 +142,44 @@ describe("HikmahCanvas expansion-notice timeout", () => {
     // Unmounting must clear the scheduled dismiss rather than letting it fire later.
     expect(clearTimeoutSpy).toHaveBeenCalled();
   });
+
+  it("surfaces a notice instead of silently dropping a connection when a different-kind edge already links the pair", async () => {
+    useCanvasStore.setState({
+      nodes: [verseNode("n1", "1:1"), verseNode("n2", "2:255")] as never,
+      edges: [
+        {
+          id: "edge-existing",
+          source: "n1",
+          target: "n2",
+          type: "hikmah",
+          data: { kind: "thematic", label: "existing" },
+        },
+      ] as never,
+    });
+    // The AI identifies a "root" connection to a verse already on the canvas,
+    // already linked to the source by a "thematic" edge — a different kind.
+    vi.stubGlobal(
+      "fetch",
+      vi.fn().mockResolvedValue({
+        ok: true,
+        json: () => Promise.resolve([{ ...connection("2:255"), kind: "root" }]),
+      })
+    );
+
+    render(<HikmahCanvas onSearchOpen={vi.fn()} />);
+
+    act(() => {
+      useCanvasStore.getState().setPendingExpand({ nodeId: "n1", ref: "1:1", kind: "root" });
+    });
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(400);
+    });
+
+    expect(screen.getByRole("status")).toHaveTextContent(
+      "These verses are already connected a different way."
+    );
+    // The pre-existing thematic edge is untouched — no second edge was added.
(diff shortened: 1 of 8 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and CLAUDE.md/AGENTS.md compliance, git history of the kind-aware dedupe and notice mechanism, prior PR precedent (PRs #353/#354 which built the notice/timer mechanism this reuses), and code-comment accuracy.

One minor note for transparency, not a blocking issue: the \`showExpansionNotice\` extraction incidentally gives the "expansion already running" notice a 6s auto-dismiss it didn't previously have (it used to persist until overwritten). Harmless/arguably-beneficial side effect of the refactor, but the PR description's "behavior is otherwise unchanged" undersells it slightly.

PR #416, opened 2026-08-10
fix(canvas): flush pending debounced autosave before page unload

### LINKED ISSUE #371: fix(canvas): debounced autosave isn't flushed before navigate-away, losing the last edit
## Summary

An edit made on the canvas within 800ms of navigating away (e.g. clicking "Sign in to save") can be lost, because the debounced `localStorage` autosave is never force-flushed before unload.

## Evidence

`hooks/useCanvasPersistence.ts:125-143` — the autosave to `localStorage` is debounced 800ms. There is no `beforeunload`/`pagehide` handler anywhere in the codebase (verified via grep across `hooks/`, `components/canvas/`, `store/`) that force-flushes the pending save.

Reproduction: make an edit on the canvas (add/move a node), then within 800ms click "Sign in to save" (`components/canvas/CanvasToolbar.tsx:155-169`), which does a hard redirect via `window.location.href = url` (line 163). The page unloads before the pending `setTimeout` fires, so the edit never reaches `localStorage`. After OAuth completes, `mergeGuestWorkspace` (`hooks/useCanvasPersistence.ts:39-66`) reads the stale pre-edit snapshot and persists that to the server.

## Impact

The user's most recent edit before signing in is silently lost, with no indication it happened.

## Suggested fix

Add a flush-on-unload path: either a `beforeunload`/`pagehide` listener that synchronously writes any pending debounced state to `localStorage`, or flush synchronously in the "Sign in to save" click handler itself before navigating.

## Acceptance criteria

- Making an edit and immediately clicking "Sign in to save" preserves that edit in the post-login merged workspace.

nazarli-shabnam:
## Summary

- The canvas autosave to `localStorage` (`hooks/useCanvasPersistence.ts`) is debounced 800ms with no flush-on-unload path. An edit made right before a hard navigation (e.g. clicking "Sign in to save", which does `window.location.href = url`) could be lost because the page unloads before the pending `setTimeout` fires.
- Added `pagehide`/`beforeunload` listeners that synchronously flush any pending debounced write using the latest node/edge state (tracked via a ref kept in sync each render).
- Extracted the save logic into a shared `saveToLocalStorage` helper used by both the debounce timeout and the unload flush, so there's one code path for the actual write.

Closes #371

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run format:check` passes locally (on changed files)
- [x] New tests added — `__tests__/hooks/useCanvasPersistence.flush.test.ts` covers: no write before debounce elapses, flush-on-pagehide, flush-on-beforeunload, and no double-write when the debounce timer fires after an unload flush already ran

## Checklist

- [x] Conventional commits
- [x] `.env.example` updated if new environment variables were added — N/A
- [x] `README.md` updated if the feature changes the public interface — N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved canvas persistence when leaving or closing the page.
  * Pending canvas edits are now saved immediately during page unload events.
  * Prevented delayed saves from running again after edits have already been flushed.
  * Preserved empty-canvas cleanup and storage error handling.

### DIFF
--- a/__tests__/hooks/useCanvasPersistence.flush.test.ts
+++ b/__tests__/hooks/useCanvasPersistence.flush.test.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { renderHook, act } from "@testing-library/react";
+
+vi.mock("@xyflow/react", () => ({
+  applyNodeChanges: vi.fn((changes: unknown[], nodes: unknown[]) => nodes),
+  applyEdgeChanges: vi.fn((changes: unknown[], edges: unknown[]) => edges),
+}));
+
+import { useCanvasStore } from "@/store/canvas";
+import { useCanvasPersistence, CANVAS_STORAGE_KEY } from "@/hooks/useCanvasPersistence";
+import type { Verse } from "@/types/quran";
+
+const baseVerse: Verse = {
+  surah: 2,
+  ayah: 255,
+  ref: "2:255",
+  arabicText: "اللَّهُ لَا إِلَٰهَ إِلَّا هُوَ",
+  translation: "Allah — there is no deity except Him.",
+  surahName: "Al-Baqarah",
+  surahNameArabic: "البقرة",
+};
+
+describe("useCanvasPersistence flush-on-unload", () => {
+  beforeEach(() => {
+    useCanvasStore.getState().reset();
+    localStorage.clear();
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("does not persist an edit before the 800ms debounce elapses", () => {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Pushed a follow-up commit after re-examining this myself: the unload flush could wipe a real localStorage canvas if it fired while a `?share=<id>` restore was still in flight (the debounced-save effect schedules a save of the still-empty initial store state before the async restore resolves). Added a `restoringRef` guard to suppress the debounced save and the unload flush during that window, dropped the `beforeunload` listener in favor of `pagehide` alone (avoids disqualifying the page from bfcache), and added a regression test covering the in-flight-restore case.

PR #421, opened 2026-08-11
fix(canvas): move first-visit tour to top-right so it doesn't overlap the legend

### LINKED ISSUE #375: fix(canvas): first-visit tour card overlaps the connection-kind legend
## Summary

The first-visit onboarding tour on the canvas page can render directly on top of the connection-kind legend, hiding it entirely.

## Evidence

`components/canvas/CanvasTour.tsx:67` positions the tour card at `md:bottom-6 md:left-6`. `components/canvas/HikmahCanvas.tsx:353-358` renders `<Panel position="bottom-left"><CanvasLegend/></Panel>` in the same corner (xyflow's default panel margin is ~15px) as soon as `nodes.length > 0`. `CanvasTour` mounts regardless of node count and is only dismissed by explicit user action (Escape, Skip, or stepping through to "Got it").

## Impact

A first-time visitor who adds a verse from the empty-state search before finishing/dismissing the tour sees the ~320-360px-wide, opaque (`z-50`) tour card rendered directly on top of the legend, hiding it completely.

## Suggested fix

Either reposition the tour card to avoid the bottom-left corner, or hide/reposition `CanvasLegend` while the tour is active.

## Acceptance criteria

- Adding a verse to the canvas while the first-visit tour is still active does not visually hide the connection-kind legend.

nazarli-shabnam:
## Summary

On desktop, the first-visit onboarding tour card (`CanvasTour`) was anchored `bottom-6 left-6`, the same corner as the connection-kind legend panel (`CanvasLegend`, rendered via `<Panel position="bottom-left">`). Since the legend only renders once `nodes.length > 0`, a new visitor who adds their first verse during the tour would see the tour card overlapping the legend.

## Fix

Moved the tour's desktop anchor to `top-right` (`md:right-6 md:top-6`), a corner nothing else occupies (the minimap defaults to bottom-right but the tour no longer competes with it since they're on different rows). The mobile position (bottom-center, above the mobile toolbar) is unchanged — the legend is hidden below the `sm` breakpoint (`hidden ... sm:flex`), so there's no overlap on mobile to begin with.

## Testing

Added a test asserting the container's desktop classes place it top-right (`md:right-6`, `md:top-6`) and not bottom-left (`md:bottom-6`, `md:left-6`). Verified the new test fails against the pre-fix code and passes with the fix. Full suite (`bun run test:ci`), lint, and typecheck all pass.

Closes #375

## Summary by CodeRabbit

* **Improvements**
  * Repositioned the Canvas tour on desktop from the lower-left corner to the upper-right.
  * Preserved the existing mobile placement.

### DIFF
--- a/__tests__/components/canvas/CanvasTour.test.tsx
+++ b/__tests__/components/canvas/CanvasTour.test.tsx
@@ -49,4 +49,15 @@ describe("CanvasTour", () => {
 
     document.body.removeChild(outside);
   });
+
+  it("positions in the top-right on desktop, not bottom-left where the canvas legend sits", async () => {
+    renderWithIntl(<CanvasTour />);
+    const dialog = await screen.findByRole("dialog");
+    const container = dialog.parentElement;
+
+    expect(container?.className).toMatch(/md:right-6/);
+    expect(container?.className).toMatch(/md:top-6/);
+    expect(container?.className).not.toMatch(/md:bottom-6/);
+    expect(container?.className).not.toMatch(/md:left-6/);
+  });
 });
--- a/__tests__/components/canvas/CanvasTour.test.tsx
+++ b/__tests__/components/canvas/CanvasTour.test.tsx
@@ -50,14 +50,14 @@ describe("CanvasTour", () => {
     document.body.removeChild(outside);
   });
 
-  it("positions in the top-right on desktop, not bottom-left where the canvas legend sits", async () => {
+  it("positions in the top-left on desktop, clear of the legend (bottom-left) and context sidebar (right edge)", async () => {
     renderWithIntl(<CanvasTour />);
     const dialog = await screen.findByRole("dialog");
     const container = dialog.parentElement;
 
-    expect(container?.className).toMatch(/md:right-6/);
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (fixed in a follow-up commit before this comment was posted):

1. The top-right position traded one overlap for another: `ContextSidebar` docks to the full right edge (`top-0 right-0 z-40 h-full`) whenever a node/edge is clicked, and the tour card (`z-50`, up to 360px wide) at `top-6 right-6` would render on top of the sidebar's header/close button — a collision that's arguably *more* likely to occur than the original legend overlap, since clicking a node is exactly what the tour's own second step encourages.

https://github.com/OpenHikmah/openhikmah-web/blob/0e8acf013ae621d9af00bf31a45081e49fc183c1/components/layout/ContextSidebar.tsx#L280-L297

Fixed in [0130f43](https://github.com/OpenHikmah/openhikmah-web/commit/0130f43) by moving the tour to top-left instead — the one corner nothing else claims (toolbar is top-center, legend is bottom-left, minimap defaults to bottom-right, sidebar is the full right edge).

PR #423, opened 2026-08-11
fix(a11y): trap focus, handle Escape, and fix tab order in mobile More sheet

### LINKED ISSUE #377: fix(a11y): mobile "More" sheet has no focus trap, no Escape handling, and broken tab order
## Summary

The mobile header's "More" sheet (Save/Export PNG/Export PDF) is effectively unusable for keyboard-only users: it isn't reachable via forward Tab after opening, traps no focus, and has no Escape-to-close handling.

## Evidence

`components/layout/Header.tsx:76-175, 350-392`. `MoreSheet` is rendered conditionally *before* the bottom action-bar `<div>` in the JSX (line 353 vs. 366), so its buttons appear **earlier** in DOM/tab order than the "More" trigger button that opens it (inside the bar at line 383). A keyboard user who Tabs to "More" and activates it with Enter/Space cannot then Tab forward into the sheet — forward Tab skips past it entirely to the "Clear" button, since the sheet's buttons are already behind the trigger in tab order. Reaching the sheet requires Shift+Tab backwards past several other buttons, which is not discoverable.

Additionally: no focus is moved into the sheet on open (no autofocus/`FocusScope`, unlike `ExpandMenu.tsx`/`ExportMenu.tsx`/`CanvasTour.tsx`, which all correctly use Radix `FocusScope trapped`), and the outside-click-to-close handler only listens for `mousedown` (lines 104-116) — there is no `keydown`/Escape handler at all.

## Impact

Keyboard-only mobile-nav users effectively cannot use Save/Export from the mobile bar without blind Shift+Tab navigation, and have no way to close the sheet via keyboard once open.

## Suggested fix

Adopt the same Radix `FocusScope trapped` pattern already used by `ExpandMenu`/`ExportMenu`/`CanvasTour` in this codebase: move focus into the sheet on open, trap it while open, add an Escape handler, and fix DOM order so forward Tab reaches the sheet naturally.

## Acceptance criteria

- Opening the "More" sheet via keyboard moves focus into it.
- Escape closes the sheet and returns focus to the trigger.
- Forward Tab from the trigger reaches the sheet's contents.

nazarli-shabnam:
## Summary

The mobile header's "More" sheet (Save/Export PNG/Export PDF) had no focus trap, no Escape handling, and was positioned before the bottom action bar in the JSX — putting its buttons behind the "More" trigger in DOM/tab order. A keyboard user who Tabbed to "More" and activated it with Enter/Space could not then Tab forward into the sheet, and had no way to close it via keyboard.

## Fix

Adopted the same Radix `FocusScope trapped loop` pattern already used by `ExpandMenu`, `ExportMenu`, and `CanvasTour` in this codebase:
- Wrapped the sheet in `FocusScope`, moving focus into it on open (Radix's default auto-focus of the first tabbable element) and returning focus to the trigger on close.
- Added an `Escape` keydown handler matching `ExportMenu`'s existing pattern.
- Moved the sheet's JSX after the bottom bar so it no longer sits ahead of its own trigger in document order.

## Testing

Added `__tests__/components/layout/Header.more-sheet-a11y.test.tsx`, covering: focus moves into the sheet on open; Tab traps/loops within the sheet (doesn't leak to the bar behind it); Escape closes the sheet and returns focus to the trigger. Verified all three fail against the pre-fix code and pass with the fix. Full suite, lint, and typecheck all pass.

Closes #377

### DIFF
--- a/__tests__/components/layout/Header.more-sheet-a11y.test.tsx
+++ b/__tests__/components/layout/Header.more-sheet-a11y.test.tsx
@@ -1,0 +1,74 @@
+import { screen, fireEvent, act } from "@testing-library/react";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+import { describe, expect, it, beforeEach, vi } from "vitest";
+
+vi.mock("next/navigation", () => ({
+  usePathname: () => "/canvas",
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn(), refresh: vi.fn() }),
+}));
+
+import { Header } from "@/components/layout/Header";
+import { useCanvasStore } from "@/store/canvas";
+import type { Verse } from "@/types/quran";
+
+const baseVerse: Verse = {
+  surah: 2,
+  ayah: 255,
+  ref: "2:255",
+  arabicText: "اللَّهُ لَا إِلَٰهَ إِلَّا هُوَ",
+  translation: "Allah — there is no deity except Him.",
+  surahName: "Al-Baqarah",
+  surahNameArabic: "البقرة",
+};
+
+beforeEach(() => {
+  useCanvasStore.getState().reset();
+  useCanvasStore.getState().addVerseNode(baseVerse, { x: 0, y: 0 });
+});
+
+describe("Header mobile More sheet — keyboard accessibility", () => {
+  it("moves focus into the sheet when it opens", async () => {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and CLAUDE.md/AGENTS.md compliance.

PR #427, opened 2026-08-11
fix(search): abort the result-select fetch when the dialog closes

### LINKED ISSUE #381: fix(search): result-select fetch isn't aborted on dialog close, can add a canvas node after dismiss
## Summary

Selecting a search result and immediately closing the search dialog can still silently add a node to the canvas and pan the camera to it, after the user believed they'd cancelled the action.

## Evidence

`components/search/SearchDialog.tsx:167-182, 306-318`. The debounced search/preview requests are tracked via `abortRef` and correctly cancelled in the `Dialog.Root` `onOpenChange` handler (lines 199-215), but the two "select this verse" fetches — `SearchResultRow`'s `onSelect` handler (lines 306-318) and `loadSeedVerse` (lines 167-182) — use bare `fetch()` calls with no `AbortSignal` at all.

Reproduction: click a search result, then immediately close the dialog (Escape or the X) before the `/api/verse/...` fetch resolves. The dialog closes and resets its own state, but the pending fetch still resolves afterward and calls `mapConnections(verse)`, adding a verse node to the canvas and panning the camera to it.

## Impact

A visible, surprising side effect (a new node appearing on the canvas) that happens after the user believed they'd cancelled the action.

## Suggested fix

Wire the same `abortRef`/`AbortSignal` already used for search/preview requests into these two "select result" fetches, and check whether the dialog is still open before acting on the result.

## Acceptance criteria

- Selecting a result and immediately closing the dialog before the fetch resolves does not add a node to the canvas.

nazarli-shabnam:
## Summary

Selecting a search result and immediately closing the dialog before the fetch resolved would still silently add a node to the canvas and pan the camera to it — a visible, surprising side effect after the user believed they'd cancelled the action.

The debounced search/preview requests in `SearchDialog.tsx` are correctly cancelled via `abortRef` in the `onOpenChange` close handler, but `selectResult` and `loadSeedVerse` used bare `fetch()` calls with no `AbortSignal` at all.

## Fix

Added a separate `selectAbortRef` (kept independent from the search/preview `abortRef` so unrelated typing in the input doesn't cancel an in-flight result selection), wired into both `selectResult` and `loadSeedVerse`, and aborted alongside the existing cleanup in the dialog's close handler. An `AbortError` is treated as expected cancellation, not a failure to surface via `selectError`.

## Testing

Added `__tests__/components/search/SearchDialog.select-abort.test.tsx`: selects a result, closes the dialog (Escape) before the `/api/verse/...` fetch resolves, then lets the network response "arrive" afterward — asserts no canvas node was added. Verified the test fails against the pre-fix code (the node is added) and passes with the fix. Full suite, lint, and typecheck all pass.

Closes #381

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved search dialog behavior when closing during an in-progress verse selection.
  - Prevented canceled requests from adding unwanted content or incorrectly changing loading states.
  - Closing the dialog now reliably cancels active search and preview requests.

- **Tests**
  - Added coverage for canceling a verse fetch by closing the dialog before the response completes.

### DIFF
--- a/__tests__/components/search/SearchDialog.select-abort.test.tsx
+++ b/__tests__/components/search/SearchDialog.select-abort.test.tsx
@@ -1,0 +1,111 @@
+import { useState } from "react";
+import { screen, fireEvent, waitFor } from "@testing-library/react";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+import type { Verse, SearchResult, SearchResponse } from "@/types/quran";
+
+vi.mock("@xyflow/react", () => ({
+  applyNodeChanges: vi.fn((changes: unknown[], nodes: unknown[]) => nodes),
+  applyEdgeChanges: vi.fn((changes: unknown[], edges: unknown[]) => edges),
+}));
+
+vi.mock("next/navigation", () => ({ useRouter: () => ({ push: vi.fn() }) }));
+
+import { useCanvasStore } from "@/store/canvas";
+import { SearchDialog } from "@/components/search/SearchDialog";
+
+const VERSE_TEXT: Record<string, { arabic: string; translation: string }> = {
+  "2:1": { arabic: "الم", translation: "Alif, Lam, Meem." },
+  "2:2": {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review

PR #428, opened 2026-08-11
fix(search): prevent debounced navigation from clobbering a direct one

### LINKED ISSUE #382: fix(search): debounced navigation can clobber an example-chip click on the search page
## Summary

Typing a query, then clicking an example-search chip before the query's debounce fires, can result in the URL/search silently reverting back to the originally-typed (unwanted) query a moment later.

## Evidence

`app/search/SearchPageClient.tsx:130-134, 161-178`. `onInputChange` starts a 400ms debounce (`debounceRef`) that calls `navigate(value, mode, 1)`. That timer is only ever cleared on component unmount (lines 70-74) or by a subsequent keystroke — clicking one of the `EXAMPLE_SEARCHES` chips calls `navigate(...)` directly and never touches `debounceRef`. The example chips are rendered whenever `!q.trim()`, i.e. before the debounced navigation has committed a `q` to the URL.

Reproduction: type a query, then click an example chip while the debounce is still pending. The chip's `navigate()` call runs first, then ~400ms later the stale debounced call fires and overwrites the URL/search back to the originally-typed query.

## Impact

A confusing, silent state reversion — the user sees the example-chip results briefly, then the page reverts to the query they'd already moved past.

## Suggested fix

Clear `debounceRef`'s pending timer whenever an example chip (or any other direct `navigate()` call) fires.

## Acceptance criteria

- Typing a query then clicking an example chip before the debounce elapses results in the chip's query staying active, not reverting.

nazarli-shabnam:
## Summary

Typing a query, then clicking an example-search chip before the query's 400ms debounce fires, could result in the URL/search silently reverting back to the originally-typed (unwanted) query a moment later.

## Evidence

`app/search/SearchPageClient.tsx` — `onInputChange` starts a 400ms debounce (`debounceRef`) that calls `navigate(value, mode, 1)`. That timer was only ever cleared on component unmount or by a subsequent keystroke — clicking one of the `EXAMPLE_SEARCHES` chips (or toggling search mode, or paging) called `navigate(...)` directly and never touched `debounceRef`.

Reproduction: type a query, then click an example chip (or the mode toggle) while the debounce is still pending. The direct `navigate()` call runs first, then ~400ms later the stale debounced call fires and overwrites the URL/search back to the originally-typed query.

## Fix

`navigate()` now clears any pending `debounceRef` timer before doing its own navigation, so a direct call (example chip, mode toggle, pagination) always wins over a still-pending debounced call from earlier typing.

## Testing

Added a regression test in `__tests__/app/search/SearchPageClient.test.tsx`: types a query, clicks an example chip before the debounce elapses, advances fake timers past the debounce window, and asserts `router.replace` was called exactly once — with the chip's query, not the typed one. Verified the test fails against the pre-fix code (2 calls, the stale one wins) and passes with the fix. Full suite, lint, typecheck, and integration tests all pass.

Closes #382

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed search navigation so direct actions, such as selecting an example or changing pages, no longer trigger outdated pending search requests.
  * Improved pagination behavior by ensuring navigation callbacks run consistently for previous, next, and numbered page links.

### DIFF
--- a/__tests__/app/search/SearchPageClient.test.tsx
+++ b/__tests__/app/search/SearchPageClient.test.tsx
@@ -140,22 +140,16 @@ describe("SearchPageClient — direct navigation clears a pending debounced navi
   it("clicking an example chip before the debounce fires wins, not the stale typed query", async () => {
     renderWithIntl(<SearchPageClient />);
 
-    // Type a query — starts the 400ms debounce that would otherwise call
-    // navigate("Patience search term", ...) once it fires.
     fireEvent.change(screen.getByRole("textbox", { name: /search verses/i }), {
       target: { value: "some other query" },
     });
 
-    // Click an example chip before the debounce elapses.
     fireEvent.click(screen.getByRole("button", { name: "Mercy" }));
 
-    // Let the original 400ms debounce timer (if not cleared) fire.
     await act(async () => {
       vi.advanceTimersByTime(500);
     });
 
-    // Only the chip's direct navigate() call should have gone through — the
-    // stale debounced call for the typed query must not have fired at all.
     expect(mockReplace).toHaveBeenCalledTimes(1);
     expect(mockReplace).toHaveBeenCalledWith(
       expect.stringContaining(`q=${encodeURIComponent("Mercy")}`)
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
@coderabbitai review please

PR #430, opened 2026-08-11
fix(social): header polling badges no longer reset to zero on failure

### LINKED ISSUE #384: fix(social): header polling badges reset to zero on a transient network failure
## Summary

The friends/mentions unread badges in the header can silently disappear on a single transient network failure, even though the user still has pending items — indistinguishable from a legitimate "nothing pending" state.

## Evidence

`components/layout/Header.tsx:225-236, 247-253`:

```js
fetch("/api/social/friends?limit=200", {...})
  .then((r) => (r.ok ? r.json() : { items: [] }))
  ...
fetch("/api/social/mentions?limit=1", {...})
  .then((r) => (r.ok ? r.json() : { unreadCount: 0 }))
```

Both 60-second polls fall back to an empty/zero result on any non-OK response.

## Impact

Reproduction: with pending friend requests or unread mentions already showing in the header badge, have one polling request return a transient 5xx/network hiccup — the badge silently drops to 0/disappears until the next successful poll, even though the user still has unread items.

## Suggested fix

On a failed poll, keep the previous known count/state instead of falling back to empty/zero — only update the badge from a successful response.

## Acceptance criteria

- A single failed poll does not clear an already-displayed non-zero badge count.

nazarli-shabnam:
## Summary

The friends/mentions unread badges in the header silently reset to zero on a single transient network failure, even though the user still has real pending items — indistinguishable from a legitimate "nothing pending" state.

## Evidence

`components/layout/Header.tsx`: both 60-second polls (friend requests, mentions) fell back to an empty/zero result on any non-OK response:

```js
fetch("/api/social/friends?limit=200", {...})
  .then((r) => (r.ok ? r.json() : { items: [] }))
  ...
fetch("/api/social/mentions?limit=1", {...})
  .then((r) => (r.ok ? r.json() : { unreadCount: 0 }))
```

## Fix

Both polls now resolve to `null` on a non-OK response and skip the state update entirely in that case, keeping the previously-known count until the next successful poll.

## Testing

Added `__tests__/components/layout/Header.polling-failure.test.tsx`, covering both the friend-request and mentions polls: a badge with a non-zero count survives one failed poll 60s later. Verified both tests fail without the fix and pass with it. `bun run lint`, `bun run typecheck`, `bun run format:check`, and the full unit suite pass.

Closes #384

## Summary by CodeRabbit

* **Bug Fixes**
  * Friend-request and unread-mention badges now retain their previous counts when polling encounters an error.
  * Badge counts continue updating normally after successful polling responses.
* **Tests**
  * Added coverage to verify badge counts remain visible when a scheduled polling request fails.

### DIFF
--- a/__tests__/components/layout/Header.polling-failure.test.tsx
+++ b/__tests__/components/layout/Header.polling-failure.test.tsx
@@ -1,0 +1,101 @@
+import { act } from "@testing-library/react";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
+
+vi.mock("next/navigation", () => ({
+  usePathname: () => "/canvas",
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn(), refresh: vi.fn() }),
+}));
+
+import { Header } from "@/components/layout/Header";
+import { useAuthStore } from "@/store/auth";
+import { useSocialStore } from "@/store/social";
+
+describe("Header — polling badges survive a transient failure", () => {
+  const mockFetch = vi.fn();
+  const previousAuth = useAuthStore.getState();
+  const previousSocial = useSocialStore.getState();
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.stubGlobal("fetch", mockFetch);
+    mockFetch.mockReset();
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review pls

PR #431, opened 2026-08-11
fix(social): mentions mark-as-read no longer clears badge on failure

### LINKED ISSUE #385: fix(social): mentions "mark as read" clears the badge even when the server request fails
## Summary

The mentions page clears the unread badge to 0 client-side even when the server-side "mark as read" request fails, causing a client/server state desync with no error surfaced.

## Evidence

`app/mentions/page.tsx:50-57`:

```js
fetch("/api/social/mentions", { method: "PATCH", ... })
  .then(() => { if (!cancelled) setPendingMentionCount(0); })
  .catch((e) => console.error("mentions page: mark-read failed", e));
```

`fetch()` only rejects on network failure, not on HTTP error status, so the `.then()` runs (and clears the badge to 0) even if the PATCH returns 4xx/5xx — there's no `r.ok` check.

## Impact

The mismatch self-heals on the next 60s poll in `Header.tsx`, but in the interim the user sees "0 mentions" in the account menu while the server still has them marked unread — a genuine desync with no error surfaced to the user.

## Suggested fix

Check `r.ok` before clearing the local count; on failure, keep the previous count and surface an error.

## Acceptance criteria

- A failed mark-as-read PATCH does not clear the locally-displayed mention count.

nazarli-shabnam:
## Summary

The mentions page cleared the unread badge to 0 client-side even when the server-side "mark as read" request failed, causing a client/server state desync with no error surfaced.

## Evidence

`app/mentions/page.tsx` (mark-as-read PATCH):

```js
fetch("/api/social/mentions", { method: "PATCH", ... })
  .then(() => { if (!cancelled) setPendingMentionCount(0); })
  .catch((e) => console.error("mentions page: mark-read failed", e));
```

`fetch()` only rejects on network failure, not on HTTP error status, so the `.then()` ran (and cleared the badge to 0) even if the PATCH returned 4xx/5xx.

## Fix

Check `r.ok` before clearing the local count. On failure (either a non-OK response or a network rejection), the previous count is kept and an inline error message is shown ("Couldn't mark mentions as read. The unread count may be out of sync.").

## Testing

Added `__tests__/app/mentions/page.test.tsx` with two tests: a failed PATCH keeps the count and shows the error message; a successful PATCH clears the count with no error shown. Verified the failure-path test fails without the fix. `bun run lint`, `bun run typecheck`, `bun run format:check`, and the full unit suite (965 tests) pass.

Closes #385

### DIFF
--- a/__tests__/app/mentions/page.test.tsx
+++ b/__tests__/app/mentions/page.test.tsx
@@ -1,0 +1,66 @@
+import { screen, act } from "@testing-library/react";
+import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn(), refresh: vi.fn() }),
+}));
+
+vi.mock("@/components/layout/LandingHeader", () => ({
+  LandingHeader: () => <div data-testid="landing-header" />,
+}));
+
+import MentionsPage from "@/app/mentions/page";
+import { useAuthStore } from "@/store/auth";
+import { useSocialStore } from "@/store/social";
+
+describe("MentionsPage — mark-as-read failure handling", () => {
+  const mockFetch = vi.fn();
+  const previousAuth = useAuthStore.getState();
+  const previousSocial = useSocialStore.getState();
+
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review pls

nazarli-shabnam:
@coderabbitai review pls

PR #432, opened 2026-08-11
fix(bookmarks): add per-ref in-flight guard to toggleBookmark

### LINKED ISSUE #386: fix(bookmarks): toggleBookmark has no in-flight guard, allowing desynced double-clicks
## Summary

Rapidly double-clicking a bookmark toggle can fire overlapping POST and DELETE requests that race each other, leaving the client-displayed bookmark state out of sync with the server.

## Evidence

`store/auth.ts:80-119` (`toggleBookmark`), invoked from unguarded `IconButton`s in `app/search/SearchPageClient.tsx:282` and `app/bookmarks/page.tsx:133`. `toggleBookmark` optimistically flips local state and fires a POST or DELETE based on the bookmark state read at call time, with no debounce, no `busy` flag, and none of the calling buttons pass a `disabled` prop while a request is in flight.

Reproduction: double-click the bookmark icon fast enough that the second click's `wasBookmarked` check reads the pre-first-click-response state — this fires both a POST and a DELETE to `/api/bookmarks/*` nearly simultaneously; whichever response lands last determines the final server state, which can end up out of sync with what's displayed locally until the next full reload/`loadRemoteBookmarks()`.

## Impact

UI can show "bookmarked" while the server actually has it removed (or vice versa), until the next full data reload.

## Suggested fix

Add an in-flight guard (per-verse `busy` flag) that disables the toggle button while a request for that item is pending, matching patterns used elsewhere in the app (e.g. `busy` state on admin action buttons).

## Acceptance criteria

- Rapidly double-clicking a bookmark toggle results in exactly one net state change, with client and server state matching afterward.

nazarli-shabnam:
## Summary

Rapidly double-clicking a bookmark toggle could fire overlapping POST and DELETE requests that race each other, leaving the client-displayed bookmark state out of sync with the server.

## Evidence

`store/auth.ts`'s `toggleBookmark`, invoked from unguarded `IconButton`s in `app/search/SearchPageClient.tsx` and `app/bookmarks/page.tsx`. It optimistically flips local state and fires a POST or DELETE based on the bookmark state read at call time, with no debounce, no busy flag, and neither caller passed `disabled` while a request was in flight.

Reproduction: double-click the bookmark icon fast enough that the second click's `wasBookmarked` check reads the pre-first-click-response state — this fires both a POST and a DELETE nearly simultaneously; whichever response lands last determines the final server state.

## Fix

Added a per-ref `bookmarkBusy: Record<string, boolean>` map to the auth store, set on the ref when a request starts and cleared in `.finally()` when it settles. `toggleBookmark` now no-ops if a request for that ref is already in flight. Both callers (`SearchPageClient.tsx`'s `ResultCard`, `bookmarks/page.tsx`) now disable the toggle button while its ref is busy.

## Testing

Added three tests to `__tests__/store/auth.test.ts`: a second toggle for the same ref while one is in flight is a no-op (no second fetch, state unchanged); the busy flag clears once the request settles, allowing a later toggle; the guard is per-ref, not global (two different refs can toggle concurrently). Verified the new tests fail without the fix. `bun run lint`, `bun run typecheck`, `bun run format:check`, and the full unit suite (966 tests) pass.

Closes #386

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate bookmark actions while a request is in progress.
  * Bookmark controls now remain disabled only for the affected verse.
  * Busy states clear after successful or failed bookmark operations.
  * Users can continue bookmarking different verses concurrently.
  * Prevented outdated bookmark requests from affecting a new session after sign-out.

### DIFF
--- a/__tests__/app/bookmarks/page.test.tsx
+++ b/__tests__/app/bookmarks/page.test.tsx
@@ -1,0 +1,64 @@
+import { screen, act, waitFor } from "@testing-library/react";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+
+vi.mock("next/navigation", () => ({
+  usePathname: () => "/bookmarks",
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn() }),
+}));
+
+import BookmarksPage from "@/app/bookmarks/page";
+import { useAuthStore } from "@/store/auth";
+import { TooltipProvider } from "@/components/ui";
+
+describe("BookmarksPage — bookmark button busy state", () => {
+  const mockFetch = vi.fn();
+  const previousAuthState = useAuthStore.getState();
+
+  beforeEach(() => {
+    vi.stubGlobal("fetch", mockFetch);
+    mockFetch.mockReset();
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    useAuthStore.setState(previousAuthState);
+  });
+
+  it("disables the remove-bookmark button while its ref is busy", async () => {
+    useAuthStore.setState({
+      bookmarks: ["2:255"],
+      bookmarksLoadError: false,
+      bookmarkBusy: { "2:255": true },
+    });
+
+    mockFetch.mockResolvedValueOnce(
+      new Response(
+        JSON.stringify({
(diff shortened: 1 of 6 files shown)


nazarli-shabnam:
@coderabbitai review pls

PR #452, opened 2026-08-12
fix(canvas): don't let a pending share-link fetch race the debounced autosave

### LINKED ISSUE #369: fix(canvas): share-link load races the debounced autosave and can wipe a guest's local canvas
## Summary

Opening a slow or broken `/canvas?share=<id>` link can silently delete a guest user's previously-saved local canvas from `localStorage`, with no error and no way to recover it.

## Evidence

`hooks/useCanvasPersistence.ts:92-143`. The mount effect (lines 93-123) sets `hydratedRef.current = true` (line 95) as its first statement, before the async share fetch resolves. React runs effects in declaration order within the same commit, so the second effect — the debounced autosave (lines 126-143) — sees `hydratedRef.current === true` on the first render and arms an 800ms timer using the still-empty initial `nodes`/`edges` (the restore hasn't landed yet).

Reproduction: open `/canvas?share=<id>` on a browser that already has a saved guest canvas in `localStorage` (`open-hikmah-canvas`), where the share fetch is slow (>800ms) or fails/returns invalid data (deleted share, network error, bad id).
- At 800ms, the autosave effect fires with `nodes.length === 0` → `localStorage.removeItem(LS_KEY)` (line 132) — the guest's saved canvas is deleted.
- The share `.catch()`/invalid-data branch then calls `tryRestoreFromLocalStorage()` (lines 112, 116) — but `raw` is now `null`, so nothing is restored.

## Impact

A guest user's previously-saved local canvas silently disappears with no error or confirmation, triggered simply by clicking a broken/slow share link.

## Suggested fix

Don't arm the autosave timer until the initial hydration (including any pending share fetch) has actually completed — e.g. use a separate `hydrationCompleteRef` set only after the share fetch (success, failure, or n/a) has resolved and any restore has run, rather than setting `hydratedRef` synchronously at mount.

## Acceptance criteria

- Opening a slow/failing share link never clears an existing guest canvas from `localStorage` before the restore/fallback path has had a chance to run.
- New test: simulate a delayed/failing share fetch with a pre-existing `localStorage` canvas and assert the stored canvas is untouched until hydration completes.

nazarli-shabnam:
## Summary

Opening a slow or broken `/canvas?share=<id>` link could silently delete a guest user's previously-saved local canvas from `localStorage`, with no error and no way to recover it.

## Root cause

`hooks/useCanvasPersistence.ts`'s mount effect set `hydratedRef.current = true` synchronously, before an in-flight `?share=<id>` fetch resolved. The debounced autosave effect (declared after it, in the same commit) gated only on `hydratedRef`/`restoringRef`, and could observe `nodes`/`edges` still in their empty initial state as "hydrated." Once its 800ms timer elapsed — while the share fetch was still slow/pending, or after it failed — it wiped the guest's real `localStorage` canvas via `saveToLocalStorage([], [])` before the share restore (or its localStorage fallback) ever ran.

## Fix

Split the single overloaded ref into two:
- `startedRef` — guards the mount effect from running its restore logic twice (e.g. React StrictMode double-invoke).
- `hydratedRef` — the real "safe to autosave" signal. It's now only flipped once the *entire* restore attempt has settled: immediately for the plain-localStorage path, or inside `.finally()` (success, invalid-data-fallback, or fetch-failure-fallback) for the `?share=` path.

The debounced autosave and unload-flush effects now gate purely on `hydratedRef`, so they can't observe a premature "hydrated" state regardless of exact promise/render interleaving — no more `restoringRef` needed.

## Test plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:ci` (982 tests, full suite)
- [x] New regression test in `__tests__/hooks/useCanvasPersistence.flush.test.ts`: pre-seeds `localStorage` with a saved canvas, opens with `?share=<id>`, advances fake timers past the 800ms autosave debounce *while the share fetch is still pending*, and asserts the pre-seeded canvas is untouched — then resolves the fetch with invalid data and asserts the fallback restore still has an intact `localStorage` entry to read from.

Closes #369

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented autosave from overwriting an existing local canvas while a shared-canvas restoration is still loading.
  * Improved restoration reliability when shared-canvas retrieval fails, allowing the saved local canvas to remain available.
  * Prevented duplicate restoration attempts during initialization, reducing potential data conflicts and preserving canvas content.

### DIFF
--- a/__tests__/hooks/useCanvasPersistence.flush.test.ts
+++ b/__tests__/hooks/useCanvasPersistence.flush.test.ts
@@ -105,4 +105,44 @@ describe("useCanvasPersistence flush-on-unload", () => {
     fetchSpy.mockRestore();
     window.history.pushState({}, "", "/canvas");
   });
+
+  it("does not wipe a previously-saved canvas when the debounced autosave timer elapses while a ?share= restore is still in flight", async () => {
+    localStorage.setItem(
+      CANVAS_STORAGE_KEY,
+      JSON.stringify({ v: 1, nodes: [{ id: "existing" }], edges: [] })
+    );
+    window.history.pushState({}, "", "/canvas?share=12345678-1234-1234-1234-123456789abc");
+
+    let resolveFetch: (value: Response) => void;
+    const fetchSpy = vi.spyOn(globalThis, "fetch").mockReturnValue(
+      new Promise((resolve) => {
+        resolveFetch = resolve;
+      })
+    );
+
+    renderHook(() => useCanvasPersistence());
+
+    // The share fetch hasn't resolved yet — advance well past the 800ms
+    // autosave debounce while nodes/edges are still the initial empty state.
+    act(() => {
+      vi.advanceTimersByTime(1000);
+    });
+
+    const saved = JSON.parse(localStorage.getItem(CANVAS_STORAGE_KEY)!);
+    expect(saved.nodes).toEqual([{ id: "existing" }]);
+
+    await act(async () => {
+      resolveFetch!({ ok: false } as Response);
+      await Promise.resolve();
+      await Promise.resolve();
+    });
+
+    // The fallback restore attempt (triggered once the fetch settles) still
+    // has an intact localStorage entry to read from.
+    const savedAfterSettle = JSON.parse(localStorage.getItem(CANVAS_STORAGE_KEY)!);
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review please

PR #453, opened 2026-08-12
fix: add pagination to the admin audit log page

### LINKED ISSUE #273: fix: admin audit log page has no pagination
## Problem
The audit log page fetches the log with no \`limit\`/offset parameter:

\`\`\`ts
const { data } = useAsync(() => api("/audit"), []);
\`\`\`

The API defaults \`limit\` to 100 (capped at 500) when none is supplied:

\`\`\`ts
const limit = Number.isInteger(limitParam) && limitParam > 0 ? Math.min(limitParam, 500) : 100;
\`\`\`

There is no "load more" control, no offset-based pagination, and no filter by action/admin/date on the page, and no indication in the UI when the returned set has been truncated.

## Impact
The audit log exists specifically so admins can investigate "every mutating admin action, newest first." Once an install accumulates more than 100 audit entries, anything older simply becomes invisible with no way to page back to it and no signal that older entries exist — a silent truncation that undermines the log's purpose during an incident investigation.

## Location
\`app/admin/audit/page.tsx:19-20\`, \`app/api/admin/audit/route.ts:12-13\`

## Suggested fix
Add offset/cursor-based "load more" pagination to the audit page (the API already supports \`limit\`, so add an \`offset\`/\`before\` cursor param), and show a hint when the page's row count equals the requested limit (implying more rows may exist).

nazarli-shabnam:
## Summary

The audit log page fetched a single page with no `limit`/`offset` param, no "load more" control, and no signal when the returned set was truncated. The audit log exists specifically so admins can investigate "every mutating admin action, newest first" — once an install accumulates more than 100 entries, anything older became silently invisible with no way to page back, undermining the log's purpose during an incident investigation.

## Changes

- `app/api/admin/audit/route.ts`: now fetches `limit + 1` rows to derive `hasMore`, and accepts an `?offset=` param (parsed the same way as the existing `?limit=`). Kept the route's own 100-default/500-cap rather than switching to the project-wide `parsePagination` helper (50 default/200 cap), since the audit log intentionally allows a larger page and the issue didn't ask to shrink it — the response shape (`limit+1`/`hasMore`) still follows the same convention as `app/api/social/friends` and `app/api/social/leaderboard`.
- `app/admin/audit/page.tsx`: tracks `entries`/`hasMore` locally (seeded from the initial `useAsync` load), and adds a "Load more" button that fetches `/audit?offset=<entries.length>` and appends — mirroring the load-more pattern already used in `app/social/page.tsx`. The button only renders when `hasMore` is true, so no more "was that all of it?" ambiguity.

## Test plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test:ci` (988 tests, full suite)
- [x] Extended `__tests__/api/admin/audit.test.ts`: offset default/clamping, and `hasMore` true/false based on row count.
- [x] New `__tests__/app/admin/audit/page.test.tsx`: initial load, "Load more" click requests the next offset and appends rows, button disappears once `hasMore` is false.

Closes #273

## Summary by CodeRabbit

- **New Features**
  - Added pagination to the admin audit log with a **Load more** option.
  - Additional entries now appear alongside existing results.
  - Audit results use consistent ordering and indicate when more entries are available.
  - Improved handling of empty results, completed pages, loading states, and errors.

- **Tests**
  - Added coverage for pagination, offsets, page limits, completion detection, ordering, and audit target rendering.

### DIFF
--- a/__tests__/api/admin/audit.test.ts
+++ b/__tests__/api/admin/audit.test.ts
@@ -7,7 +7,8 @@ vi.mock("@/lib/admin/admin-auth", () => ({ requireAdmin: vi.fn() }));
 function makeDbChain(
   resolveWith: unknown = [],
   onLimit?: (n: number) => void,
-  onOffset?: (n: number) => void
+  onOffset?: (n: number) => void,
+  onOrderBy?: (cols: unknown[]) => void
 ) {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const chain: any = new Proxy(
@@ -21,6 +22,8 @@ function makeDbChain(
             Promise.resolve(resolveWith).then(res, rej);
         if (prop === "limit" && onLimit) return (n: number) => (onLimit(n), chain);
         if (prop === "offset" && onOffset) return (n: number) => (onOffset(n), chain);
+        if (prop === "orderBy" && onOrderBy)
+          return (...cols: unknown[]) => (onOrderBy(cols), chain);
         return () => chain;
       },
       apply() {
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
@coderabbitai review pls

PR #457, opened 2026-08-12
fix(names): degrade reflection/pairings gracefully on AI provider failure instead of 500ing

### LINKED ISSUE #443: fix: divine name reflection and pairings sections return 500 on every /names/[slug] page
## Describe the bug

On every divine-name detail page (`/names/[slug]`), the "Believer's Reflection" and "Common Pairings" sections fail to load. The frontend handles it gracefully (shows "Could not load the reflection/pairings at this time"), but the underlying API endpoints are broken for all 99 names.

## Steps to reproduce

1. Go to https://openhikmah.com/names/ar-rahman (or any other name, e.g. /names/al-malik)
2. Open browser devtools console / network tab
3. Observe requests to `/api/names/[slug]/reflection` and `/api/names/[slug]/pairings`

**Expected behaviour:** Both endpoints return the AI-generated reflection and common-pairings content for the name.

**Actual behaviour:** Both return `500 Internal Server Error` with body `{"error":"Internal server error"}`. Confirmed on at least `ar-rahman` and `al-malik`, so this looks systemic rather than name-specific. The `/api/names/[slug]/verses` endpoint on the same page works fine (200).

## Environment

- Bun version (`bun -v`):

```
[ERROR] Failed to load resource: the server responded with a status of 500 () @ https://openhikmah.com/api/names/ar-rahman/reflection:0
[ERROR] Failed to load resource: the server responded with a status of 500 () @ https://openhikmah.com/api/names/ar-rahman/pairings:0
```

## Additional context

This is core theological content (AI-generated reflection + pairings per the Maturidi/Hanafi taxonomy), so it affects the primary value of the Asma'ul Husna section across the whole site, not an edge case. Found during a full-site QA pass on 2026-08-12.

nazarli-shabnam:
## Summary

- `/api/names/[slug]/reflection` and `/api/names/[slug]/pairings` returned `500 Internal Server Error` on every name in production, while `/api/names/[slug]/verses` on the same page kept working.
- Root cause: `getReflection()` and `getPairings()` called `callAI()` with no local error handling, so any AI-provider failure (rate limit, quota, transient outage) propagated straight to the route's outer `catch` → hard 500. `verses/route.ts`'s `buildReasons()`/`fallbackAIVerses()` already wrap their `callAI()` calls in try/catch and fall back to a default reason — that's why verses stayed up through the same underlying failure while reflection/pairings crashed.
- **Reproduced locally**: pointed dev at a real (but currently zero-credit) Anthropic key. Before the fix: `verses` → 200 (falls back to `"Contains a form of {name}."` reasons), `reflection`/`pairings` → 500 with the exact same underlying `invalid_request_error`. After the fix: all three return 200, with `reflection`/`pairings` degrading to empty content — matching the frontend's existing "Could not load..." fallback UI instead of surfacing a raw 500.
- Fix: wrap the `callAI()` call in both routes' generate function in the same try/catch pattern already used in `verses/route.ts`, logging and returning empty (not persisted, so the next request retries — matches the existing `isEmpty` contract in `lib/names/name-content.ts`).
- **Scope note**: this closes the code-level gap (missing error handling causing 500s instead of graceful degradation). If production's actual AI calls are failing for an account/billing/quota reason rather than a transient blip, that's an ops-side fix outside this repo's code — but it will now be visible via `console.error` logs instead of a raw 500, and once the underlying provider issue clears, the very next request regenerates real content (nothing gets stuck caching an empty result).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

**Details**: No prompt text, model, or theological framing changed — this only adds error handling around the existing `callAI()` calls so a provider failure degrades gracefully instead of crashing the route.

## Testing

- [x] `bun run test:ci` passes locally (995/995)
- [ ] `bun run format:check` passes locally — pre-existing unrelated formatting warnings in `.superpowers/sdd/*.md`, not touched by this PR
- [x] New tests added for new functionality — added regression tests in `__tests__/api/names-ai-validation.test.ts` asserting both routes return `200` with empty content (not `500`) when `callAI` rejects.

## Checklist

- [x] Conventional commits

Fixes #443

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of AI service failures for name reflections and pairings, returning graceful empty results instead of server errors.
  * Added visible error messaging when reflection content is unavailable, preventing the loading state from persisting indefinitely.
  * Added monitoring for AI call failures to improve issue detection.

* **Tests**
  * Added coverage for AI failures, fallback responses, error logging, and unavailable reflection content.

### DIFF
--- a/__tests__/api/names-ai-validation.test.ts
+++ b/__tests__/api/names-ai-validation.test.ts
@@ -169,6 +169,24 @@ describe("names AI routes — model output validation", () => {
     expect(prompt).toMatch(/strict tanzih/i);
   });
 
+  it("reflection: an AI provider failure returns 200 with empty content, not a 500", async () => {
+    mockCallAI.mockRejectedValue(new Error("Your credit balance is too low to access the API."));
+
+    const res = await getReflection(req("ar-rahman", "reflection"), params("ar-rahman"));
+
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({ reflection: "" });
+  });
+
+  it("pairings: an AI provider failure returns 200 with an empty array, not a 500", async () => {
+    mockCallAI.mockRejectedValue(new Error("Your credit balance is too low to access the API."));
+
+    const res = await getPairings(req("ar-rahman", "pairings"), params("ar-rahman"));
+
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual([]);
+  });
+
   it("reflection: no language directive for the default (English) locale", async () => {
     mockCallAI.mockResolvedValue("A reflection paragraph.");
     await getReflection(req("ar-rahman", "reflection"), params("ar-rahman"));
--- a/__tests__/api/names-ai-validation.test.ts
+++ b/__tests__/api/names-ai-validation.test.ts
@@ -169,22 +169,34 @@ describe("names AI routes — model output validation", () => {
     expect(prompt).toMatch(/strict tanzih/i);
   });
 
-  it("reflection: an AI provider failure returns 200 with empty content, not a 500", async () => {
+  it("reflection: an AI provider failure returns 200 with empty content, not a 500, and logs it", async () => {
     mockCallAI.mockRejectedValue(new Error("Your credit balance is too low to access the API."));
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
 
     const res = await getReflection(req("ar-rahman", "reflection"), params("ar-rahman"));
 
     expect(res.status).toBe(200);
     expect(await res.json()).toEqual({ reflection: "" });
+    expect(errorSpy).toHaveBeenCalledWith(
+      expect.stringContaining("Reflection: AI call failed"),
+      expect.any(Error)
+    );
+    errorSpy.mockRestore();
   });
 
-  it("pairings: an AI provider failure returns 200 with an empty array, not a 500", async () => {
+  it("pairings: an AI provider failure returns 200 with an empty array, not a 500, and logs it", async () => {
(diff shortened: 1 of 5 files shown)


nazarli-shabnam:
Pushed a follow-up commit addressing findings from a self-review pass:

- **Fixed a regression this PR introduced**: \`NameReflection.tsx\` used \`!reflection\` to decide skeleton-vs-loaded, which conflated "not yet fetched" (\`null\`) with "fetched but degraded to empty" (\`""\`) — the new case this fix's graceful-degradation introduces. A provider failure now correctly shows the existing "Could not load..." message instead of spinning forever. Added a component test for it.
- Added an \`incr("names_ai_call_error")\` metric on the AI-failure path in both routes, matching the existing \`incr()\` convention used for provider failures elsewhere in this file tree — makes a full outage alertable via \`/api/metrics\` instead of only visible by grepping logs.
- Strengthened the two new route tests to assert the failure is actually \`console.error\`-logged, not just that the response degrades gracefully (previously a deleted log line would've passed silently).

All 996 unit tests, lint, and typecheck pass locally.

nazarli-shabnam:
@coderabbitai review the PR please

nazarli-shabnam:
Verified against #443 and the route code — the missing inner try/catch around `callAI()` was the real root cause, and the fix correctly mirrors the existing tolerant pattern in `verses/route.ts`. Good catch on the follow-up `reflection === null`/`""` UI regression too. One note: CodeRabbit hit its rate limit and never completed an automated review on this one, so it's had manual scrutiny only — worth a `@coderabbitai review` retrigger if you want the bot's pass before merging. Thanks for the fix!

Issue #446, opened 2026-08-12
fix: /api/auth/refresh 401 logged as console error on every page load and navigation for logged-out users


nazarli-shabnam:
## Describe the bug

Every page load and every client-side navigation triggers a `POST /api/auth/refresh` call that returns `401` for anonymous visitors, and it's logged as a hard console error each time — not just once per session, but on every route change.

## Steps to reproduce

1. Open https://openhikmah.com in a logged-out session with devtools console open.
2. Navigate between a few pages (e.g. `/`, `/today`, `/search`, `/names`).

**Expected behaviour:** An anonymous visitor with no refresh token shouldn't cause a repeated failed network call logged as a console error on every navigation. Either skip the refresh attempt when there's no token to refresh, or log it at a lower severity (or not at all) for the expected "not logged in" case.

**Actual behaviour:** Console shows `Failed to load resource: the server responded with a status of 401 () @ https://openhikmah.com/api/auth/refresh:0` on the initial load and again after each subsequent navigation.

## Environment

- Browser / OS: Chromium (Playwright), Windows
- Bun version (`bun -v`):
- Branch / commit: main @ HEAD (production)
- Is this reproducible on [openhikmah.com](https://openhikmah.com)? [x] Yes [ ] No [ ] Not checked

## Screenshots / console errors

```
[ERROR] Failed to load resource: the server responded with a status of 401 () @ https://openhikmah.com/api/auth/refresh:0
```

## Additional context

Low severity — functionally harmless — but it's noisy repeated error-level logging for the common case of an anonymous visitor, which can mask real errors during debugging. Found during a full-site QA pass on 2026-08-12.

Issue #445, opened 2026-08-12
fix: homepage shows "Welcome back" logged-in-style dashboard while header still shows "Log in"


nazarli-shabnam:
## Describe the bug

After using Canvas (or after being redirected from a gated page, see related issue), the homepage switches from the marketing hero to a personalized-looking "Welcome back" dashboard — but the header still shows a "Log in" button. The page presents two contradictory signals about the visitor's auth state at the same time.

## Steps to reproduce

1. In a logged-out session, use the Canvas feature to add a few verses (e.g. https://openhikmah.com/canvas?verse=2:153, expand a couple of connections).
2. Navigate to https://openhikmah.com/ (or get redirected there from a gated route).
3. Observe the hero section and the header simultaneously.

**Expected behaviour:** If the "Welcome back" / "Continue your canvas" dashboard variant is meant for anonymous users with local canvas progress, its copy shouldn't imply an authenticated returning-user greeting ("Assalamu Alaykum, Welcome back") while the header simultaneously offers "Log in" — pick one signal. If it's meant only for authenticated users, it shouldn't render for a logged-out session at all.

**Actual behaviour:** Hero shows "ASSALAMU ALAYKUM / Welcome back" with a "Continue your canvas — N verses in progress" panel and a "Saved" panel, while the top-right header button still reads "Log in".

## Environment

- Browser / OS: Chromium (Playwright), Windows
- Bun version (`bun -v`):
- Branch / commit: main @ HEAD (production)
- Is this reproducible on [openhikmah.com](https://openhikmah.com)? [x] Yes [ ] No [ ] Not checked

## Screenshots / console errors

Screenshot shows "Welcome back" hero + "Continue your canvas (4 verses in progress)" + "Saved / No bookmarks yet" panels, with "Log in" button visible top-right in the same viewport.

## Additional context

Likely driven by local/session canvas state being used as a proxy for "returning user" without checking actual auth state for the greeting copy. Found during a full-site QA pass on 2026-08-12.

Issue #444, opened 2026-08-12
fix: auth-gated pages silently redirect anonymous users to home with no explanation


nazarli-shabnam:
## Describe the bug

Visiting an auth-gated route while logged out doesn't show a login prompt, an error, or any explanation — the browser is silently redirected to `/` (the homepage). There's no indication of why you ended up there or how to get back to what you were trying to view.

## Steps to reproduce

1. In a logged-out session, navigate directly to any of:
   - https://openhikmah.com/mentions
   - https://openhikmah.com/workspaces
   - https://openhikmah.com/social
   - https://openhikmah.com/onboarding
2. Observe the URL bar after the page settles.

**Expected behaviour:** Either a visible "please log in to continue" state (ideally with a link to log in and return to the intended page), or a redirect to a dedicated login page — something that tells the user what happened.

**Actual behaviour:** The URL silently changes to `https://openhikmah.com/` and the homepage renders with no message about the redirect. From the user's perspective the link/bookmark they followed just "did nothing" and dumped them on the homepage.

## Environment

- Browser / OS: Chromium (Playwright), Windows
- Bun version (`bun -v`):
- Branch / commit: main @ HEAD (production)
- Is this reproducible on [openhikmah.com](https://openhikmah.com)? [x] Yes [ ] No [ ] Not checked

## Screenshots / console errors

N/A — no console errors, this is a silent navigation/redirect issue.

## Additional context

Related to #see-also the homepage state shown after this redirect can look like a logged-in "Welcome back" dashboard while the header still shows "Log in" — filed separately since that's a distinct content/UX issue. Found during a full-site QA pass on 2026-08-12.

PR #460, opened 2026-08-12
fix(onboarding): gate redirect on isSessionLoading, not just accessToken

### LINKED ISSUE #366: fix(onboarding): redirect races session restoration, can permanently bounce new users off onboarding
## Summary

A new user who refreshes or directly navigates to `/onboarding` mid-flow can be immediately redirected away to `/` before their session has finished restoring, with no path back to complete onboarding (e.g. setting a username).

## Evidence

`app/onboarding/page.tsx:20-24`:

```js
useEffect(() => {
  if (!accessToken) router.replace("/");
}, [accessToken, router]);
if (!accessToken) return null;
```

`accessToken` starts `null` in the Zustand auth store and is only populated after `SessionRestorer`'s async `/api/auth/refresh` call resolves (`components/providers.tsx:85-132`, gated by `isSessionLoading`). Every other protected page correctly waits — e.g. `AuthShell.tsx:14-17` checks `isSessionLoading` before deciding to redirect. `app/onboarding/page.tsx` skips that check, so the effect fires while `accessToken` is still `null` on first render and immediately redirects.

The only place that ever routes a user to `/onboarding` is the OAuth callback's `isNewUser` branch (`app/callback/CallbackClient.tsx:80`), and that flag is false on subsequent logins/refreshes.

## Impact

A new user who refreshes the tab (or otherwise triggers a re-render before session restore completes) while on `/onboarding` is bounced to the homepage and has no way back to finish onboarding, with no error message explaining why.

## Suggested fix

Gate the redirect on `isSessionLoading`, matching `AuthShell.tsx`'s pattern: don't redirect while the session is still restoring, only redirect once restoration has completed and `accessToken` is confirmed absent.

## Acceptance criteria

- Refreshing `/onboarding` while a valid refresh token/session exists does not redirect the user away.
- A genuinely unauthenticated visit to `/onboarding` still redirects to `/`.

nazarli-shabnam:
**Part of the auth/session hardening cluster — targets `fix/auth-session-cluster`, not `main`.** This branch will merge into `main` together with #444, #445, #446 once all four land on the parent branch.

## Summary

- A new user who refreshes the tab (or otherwise re-renders) mid-`/onboarding` flow was permanently bounced to `/` with no way back to finish onboarding.
- Root cause: `app/onboarding/page.tsx`'s redirect effect fired on `!accessToken` alone. `accessToken` starts `null` in the Zustand auth store and is only populated after `SessionRestorer`'s async `/api/auth/refresh` call resolves — so the effect fires on that transient `null` on first render, before restoration has had a chance to complete.
- `components/layout/AuthShell.tsx` (used by every other gated page) already gates its redirect on `isSessionLoading` first. Applied the same pattern here: don't decide "no session" until `isSessionLoading` is `false`.

## Type of change

- [x] Bug fix

## AI / Theological changes

## Auth / security surface

This touches session-gating logic (not `lib/auth/`, `app/callback/`, or the PKCE flow itself), calling it out per the repo guardrails. The change is a stricter, later redirect condition (gate on `isSessionLoading` first) — it doesn't loosen any auth check; if anything it removes a false-positive "not logged in" redirect for a user who is actually still authenticated but whose session hasn't finished restoring yet.

## Testing

- [x] `bun run test:ci` passes locally (997/997)
- [ ] `bun run format:check` passes locally — pre-existing unrelated formatting warnings in `.superpowers/sdd/*.md`, not touched by this PR
- [x] New tests added — `__tests__/app/onboarding/page.test.tsx`: no redirect while `isSessionLoading`, redirects once loading finishes with genuinely no token, renders the form once a token is present, and doesn't redirect if `isSessionLoading` flips back to `true` after a token was already established.

## Checklist

- [x] Branch is up to date with `fix/auth-session-cluster` (its parent, which is up to date with `main`)
- [x] Conventional commits

Fixes #366

### DIFF
--- a/__tests__/app/onboarding/page.test.tsx
+++ b/__tests__/app/onboarding/page.test.tsx
@@ -1,0 +1,73 @@
+import { screen, act } from "@testing-library/react";
+import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
+import { renderWithIntl } from "../../test-utils/render-with-intl";
+
+const mockReplace = vi.fn();
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: vi.fn(), replace: mockReplace, refresh: vi.fn() }),
+}));
+
+import OnboardingPage from "@/app/onboarding/page";
+import { useAuthStore } from "@/store/auth";
+import { useSocialStore } from "@/store/social";
+
+describe("OnboardingPage — redirect vs. session restoration race", () => {
+  const previousAuth = useAuthStore.getState();
+  const previousSocial = useSocialStore.getState();
+
+  beforeEach(() => {
+    mockReplace.mockReset();
+  });
+
+  afterEach(() => {
+    useAuthStore.setState(previousAuth);
+    useSocialStore.setState(previousSocial);
+  });
+
+  it("does not redirect while the session is still restoring, even though accessToken is still null", async () => {
+    useAuthStore.setState({ accessToken: null, isSessionLoading: true });
+
+    await act(async () => {
+      renderWithIntl(<OnboardingPage />);
+    });
+
+    expect(mockReplace).not.toHaveBeenCalled();
+    expect(screen.queryByText("Choose a username")).not.toBeInTheDocument();
+  });
+
+  it("redirects home once the session has finished restoring and there is genuinely no token", async () => {
+    useAuthStore.setState({ accessToken: null, isSessionLoading: false });
+
+    await act(async () => {
+      renderWithIntl(<OnboardingPage />);
+    });
+
+    expect(mockReplace).toHaveBeenCalledWith("/");
+  });
+
+  it("renders the onboarding form once session restoration completes with a token — the case a refresh mid-flow must not lose", async () => {
+    useAuthStore.setState({ accessToken: "test-token", isSessionLoading: false });
+
+    await act(async () => {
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Pushed a follow-up commit addressing findings from a self-review pass:

- Replaced the "isSessionLoading flips back to true" test, which simulated an unreachable state (nothing ever sets \`isSessionLoading\` back to \`true\` once loaded — checked \`store/auth.ts\`) and couldn't fail against the pre-fix code, with a sequence-driven test that mirrors \`SessionRestorer\`'s real ordering (\`accessToken\` gets set before \`isSessionLoading\` settles to \`false\`). Mutation-verified: reverted the component to the exact pre-fix redirect logic locally, confirmed this test (and the loading-state test) fail against it, then restored the fix.
- The loading spinner had no assertion despite being the user-visible half of the fix — gave it \`role="status"\` + an \`aria-label\` and asserted on it.

All 997 unit tests, lint, and typecheck pass locally.

nazarli-shabnam:
Checked against #366 — gating the redirect on `isSessionLoading` before `!accessToken` correctly mirrors `AuthShell.tsx`'s existing pattern and fixes the race. The flaky E2E job on this PR was a transient Bun-download 503 from GitHub Releases, unrelated to the change — re-ran it and it's green now. Thanks!

PR #461, opened 2026-08-12
fix(auth): show a sign-in prompt instead of silently redirecting from gated pages

### LINKED ISSUE #444: fix: auth-gated pages silently redirect anonymous users to home with no explanation

nazarli-shabnam:
**Part of the auth/session hardening cluster — targets \`fix/auth-session-cluster\`, not \`main\`.** This branch will merge into \`main\` together with #366, #445, #446 once all four land on the parent branch.

**Depends on #366**: this branch includes #366's commits (merged in locally) since both touch \`app/onboarding/page.tsx\` and #444 builds on top of #366's \`isSessionLoading\` gate rather than the pre-#366 redirect logic. Until #366 merges into \`fix/auth-session-cluster\`, this PR's diff will show both changes combined — the onboarding-specific diff here is only the second \`if (!accessToken)\` block (the sign-in prompt) plus the import/hook changes; everything else onboarding-related is #366's.

## Summary

- Auth-gated pages (\`/mentions\`, \`/workspaces\`, \`/social\`, \`/onboarding\`) bounced anonymous visitors straight to \`/\` with no explanation — the URL just changed and the homepage rendered, with nothing telling the user why the link/bookmark they followed "did nothing."
- \`AuthShell.tsx\` (used by mentions/workspaces/social) and the onboarding page's own gate now render an inline "Sign in to continue" message with a working Log in button once session restoration has genuinely confirmed there's no token, instead of redirecting away. The URL stays on the page the user actually requested.
- Extracted the PKCE sign-in flow (\`buildAuthUrl\` + sessionStorage + redirect), previously duplicated in \`AccountMenu.tsx\` and \`CanvasToolbar.tsx\`, into \`hooks/useSignIn.ts\` so \`AuthShell\` and the onboarding page reuse the exact same flow as the header's existing Log in button rather than adding a third copy.
- Added \`authGate.title\`/\`authGate.body\` translation keys across all four locales (en/tr/ru/az).

## Scope note: post-login return path

After signing in, the user still lands on \`/\` (or \`/onboarding\` for new users) rather than being returned to the specific page they were on — \`app/callback/CallbackClient.tsx\` doesn't currently support a return-to path, and the issue itself only says "ideally" on this point. Implementing it would touch the PKCE/callback flow, which AGENTS.md flags as high-risk auth surface, and is a larger, separate change. Out of scope here since the core complaint — a silent, unexplained redirect — is fully fixed without touching that flow.

## Type of change

- [x] Bug fix

## AI / Theological changes

## Auth / security surface

Touches \`components/layout/AuthShell.tsx\` (session-gating UI) and \`app/onboarding/page.tsx\`, calling it out per the repo guardrails. **Not** touched: \`lib/auth/\`, \`app/callback/\`, or the PKCE flow itself — \`hooks/useSignIn.ts\` is a straight extraction of existing, unchanged sign-in-redirect logic (same \`buildAuthUrl\` call, same sessionStorage keys, same \`window.location.href\`), not new auth logic. The change removes an auth *decision* (redirect vs. stay) but doesn't change who is considered authenticated or loosen any check — an anonymous user still can't see gated content; they now see why instead of landing on \`/\` unexplained.

## Testing

- [x] \`bun run test:ci\` passes locally (1000/1000)
- [x] \`bun run typecheck\` passes locally
- [x] \`bun run lint\` passes locally
- [ ] \`bun run format:check\` passes locally — pre-existing unrelated formatting warnings in \`.superpowers/sdd/*.md\`, not touched by this PR
- [x] New tests added — \`__tests__/components/layout/AuthShell.test.tsx\` (loading/authenticated/sign-in-prompt states, no redirect call in any case) and updated \`__tests__/app/onboarding/page.test.tsx\` for the new sign-in-prompt branch.
- [ ] **Could not verify visually in a live browser in this environment** (Playwright MCP was unavailable this session) — relied on the component tests above instead. Worth a manual check of \`/mentions\`, \`/workspaces\`, \`/social\`, and \`/onboarding\` while logged out before merging.

## Checklist

- [x] Branch is up to date with \`fix/auth-session-cluster\` (its parent, which is up to date with \`main\`)
- [x] Conventional commits
- [x] \`.env.example\` updated if new environment variables were added — n/a
- [x] \`README.md\` updated if the feature changes the public interface — n/a

Fixes #444


nazarli-shabnam:
Verified live in a real browser (drove the project's own Playwright dependency directly, since the MCP browser tool was unavailable this session): all four gated pages (\`/mentions\`, \`/workspaces\`, \`/social\`, \`/onboarding\`), visited logged-out, now stay on their own URL and render the "Sign in to continue" prompt with a working Log in button — no silent redirect to \`/\`. Screenshots looked clean and consistent with the existing design language on both the header-wrapped pages and the standalone onboarding page.

nazarli-shabnam:
Pushed a follow-up commit (`3411233`) addressing findings from a self-review pass:

- **Test couldn't actually catch a reintroduced redirect.** `AuthShell.test.tsx` had no `next/navigation` mock, so a reintroduced `useRouter()`/`router.replace(...)` would throw ("expected app router to be mounted") rather than fail an assertion — masking a real regression behind an unrelated crash. Added a `next/navigation` mock with a `mockReplace` spy and `expect(mockReplace).not.toHaveBeenCalled()` in all three tests. Mutation-verified: temporarily reintroduced the old `useEffect` + `router.replace("/")` redirect into `AuthShell.tsx`, confirmed the test now fails against it, then restored the fix.
- Restored a dropped "why" comment on `CanvasToolbar.tsx`'s sign-in button (canvas persists to localStorage, so signing in and returning keeps the current graph).
- Added `console.error` logging in `useSignIn`'s catch block, per the repo's no-silent-catch guardrail.

All unit tests, lint, and typecheck pass locally.

nazarli-shabnam:
Verified against #444 — the sign-in prompt covers all four routes named in the issue (`/mentions`, `/workspaces`, `/social`, `/onboarding`), and `useSignIn.ts` is confirmed to be a straight extraction of the existing PKCE-kickoff logic (no `lib/auth/` or callback changes). The flaky Integration job was the same transient Bun-download 503 seen on the other cluster PRs — re-ran it and it's green now. Thanks for the careful auth-surface disclosure too!

PR #463, opened 2026-08-13
fix(auth): skip refresh call and its 401 for anonymous visitors

### LINKED ISSUE #446: fix: /api/auth/refresh 401 logged as console error on every page load and navigation for logged-out users

nazarli-shabnam:
**Part of the auth/session hardening cluster — targets `fix/auth-session-cluster`, not `main`.** This branch will merge into `main` together with #366, #444, #445 once all four land on the parent branch.

**Depends on #366, #444, #445**: this branch includes their commits (merged in locally), since it's cut from the tip of the last unmerged issue branch in the cluster. Until those merge into `fix/auth-session-cluster`, this PR's diff will show all four combined — the actual new work here is limited to `lib/auth/session-cookie.ts`, the `qf_has_session` cookie set/clear calls in `app/api/auth/{exchange,refresh,signout}/route.ts`, the skip-refresh check in `components/providers.tsx`'s `SessionRestorer`, and the corresponding test updates.

## Summary

- Every page load and client-side navigation triggered a `POST /api/auth/refresh` call that returns `401` for anonymous visitors, and it was logged as a hard console error each time — not just once per session, but on every route change.
- Root cause: `SessionRestorer` (`components/providers.tsx`) called `/api/auth/refresh` unconditionally on mount, with no way to know ahead of time whether there was a refresh token to redeem. The real `qf_refresh_token` cookie is `HttpOnly` (correctly, since it holds token material), so client JS has no way to check its presence before firing the request.
- Fix: a new non-`HttpOnly` marker cookie, `qf_has_session` (no token material — just a flag), is set alongside the real refresh-token cookie whenever `exchange` or `refresh` issues one, and cleared alongside it on `signout` or an invalid-refresh 401. `SessionRestorer` reads this marker and skips the `/api/auth/refresh` call entirely when it's absent, instead of firing it and eating the resulting 401.
- This implements the issue's preferred fix ("skip the refresh attempt when there's no token to refresh") rather than its fallback suggestion (lowering log severity).

## Type of change

- [x] Bug fix

## AI / Theological changes

## Auth / security surface

**Touches `lib/auth/` and all three auth routes (`exchange`, `refresh`, `signout`) — calling this out explicitly per the repo guardrails**, since this is genuine high-risk auth surface, not a superficial change.

- `lib/auth/session-cookie.ts` is new: it only defines the marker cookie's name and options (`httpOnly: false`, `secure` in production, `sameSite: lax`) — it holds no token material and cannot be used to forge a session, since the server never reads or trusts this cookie for authorization; it's a client-side UX signal only. Every actual auth decision still runs off the `HttpOnly` `qf_refresh_token` cookie exactly as before.
- The marker cookie is set/cleared in lockstep with the real refresh-token cookie in all three routes (issued together on `exchange`/successful `refresh`, cleared together on `signout` and on an invalid/expired refresh), so it can't drift out of sync and cause a false "has session" or "no session" signal.
- No change to the PKCE flow, `app/callback/`, token exchange logic, or any authorization check — this only adds a client-visible signal for the already-conditional refresh trigger.

## Testing

- [x] `bun run test:ci` passes locally (1005/1005)
- [x] `bun run test:integration` passes locally (via pre-push hook, Testcontainers)
- [x] New/updated tests — `auth-exchange.test.ts`/`auth-refresh.test.ts`/`auth-signout.test.ts` assert the marker cookie is set alongside the refresh cookie on success and cleared alongside it on signout/invalid-refresh; `providers.test.tsx` adds cases for both "no marker cookie → skip refresh" and "marker cookie present → calls refresh".

## Checklist

- [x] Branch is up to date with `fix/auth-session-cluster` (its parent, which is up to date with `main`)
- [x] Conventional commits

Fixes #446


nazarli-shabnam:
### Code review

No issues found. Checked for bugs, AGENTS.md compliance, historical git context on these files, past PR review comments, and adherence to existing in-code invariants (HttpOnly-only exposure of the real refresh token, the invalid/transient cookie-clearing split, and the redundant-refresh guard).

The new `qf_has_session` marker cookie in `lib/auth/session-cookie.ts` is set/cleared in lockstep with the real `qf_refresh_token` cookie in all three routes (same `maxAge`, paired guards), carries no token material, and is read in exactly one place (`SessionRestorer`) purely to skip an otherwise-doomed refresh call — it isn't used anywhere for an authorization decision.

PR #466, opened 2026-08-13
fix(admin): require confirmation before Create & activate on prompts

nazarli-shabnam:
## Summary

Closes #274.

On the admin prompts page, "Create & activate" pushes a brand-new, unreviewed template live for the AI connection-generation flow in one click — the insert transaction atomically deactivates the current active version (`app/api/admin/prompts/route.ts`). "Roll back to this version", which reverts to a *previously working* version, already required confirmation via `ConfirmButton`. The riskier action had no safety gate at all.

## Changes

- Wrapped "Create & activate" in the same `ConfirmButton` component (`components/admin/primitives.tsx`) already used for "Roll back to this version" on the same page.
- Kept the existing "Saving…" label during the in-flight request.
- Removed the now-unused `Button` import.

## AI / Theological changes

- [x] Changes how AI prompt content reaches production — not a change to the prompt *content* or theological framing itself, but this closes a gap where a typo'd/malformed template could go live with zero confirmation, silently affecting every subsequent connection-generation call until noticed. No wording/prompt changes in this PR.

## Test plan

- Added `__tests__/app/admin/prompts/page.test.tsx`: asserts the first click does not call the API, and only a confirming second click issues the `POST /prompts` request.
- `bun run lint`, `bun run typecheck`, `bun run test:ci` pass locally.
- `bun run test:integration` passed via the pre-push hook.

## Summary by CodeRabbit

- **New Features**
  - Added a two-step confirmation before creating and activating a prompt version.
  - The confirmation state clearly indicates the pending action with a danger-styled button.
  - The draft editor is temporarily disabled while confirmation is active or changes are being saved.

- **Tests**
  - Added coverage for confirmation behavior, request payloads, and editor state during confirmation.

Issue #274, opened 2026-07-15
fix: prompts Create & activate has no confirmation, unlike rollback


nazarli-shabnam:
## Problem
On the admin prompts page, "Create & activate" is a plain button with no confirmation step, while "Rollback to this version" is wrapped in the app's \`ConfirmButton\` pattern:

- Create & activate: plain \`<Button>\` (\`app/admin/prompts/page.tsx:117-121\`)
- Rollback: wrapped in \`ConfirmButton\` (\`app/admin/prompts/page.tsx:159-168\`)

Creating a new prompt version immediately deactivates the current active template and makes the new one live for the AI connection-generation flow — the insert transaction flips \`active\` atomically (\`app/api/admin/prompts/route.ts:60-71\`).

## Impact
The riskier of the two actions — pushing a brand-new, unreviewed template live to production in one click — has no confirmation, while reverting to a *previously working* version does. A typo'd or malformed template (e.g. a broken placeholder) goes live immediately with no "are you sure" step, and every subsequent AI connection-generation call uses it until someone notices and manually rolls back. The safety gate is inverted relative to actual risk.

## Location
\`app/admin/prompts/page.tsx:117-121\` (Create & activate) vs. \`159-168\` (Rollback)

## Suggested fix
Wrap "Create & activate" in the same \`ConfirmButton\` component already used for rollback and other high-impact actions elsewhere in the admin UI.

PR #467, opened 2026-08-13
fix(admin): require confirmation before names Save overwrites cache

### LINKED ISSUE #275: fix: names page Save overwrites cached content with no confirmation and no history
## Problem
On the admin names page, "Save" (which \`PATCH\`es \`/api/admin/names\` and overwrites the stored \`data\` payload outright) has no confirmation, while "Invalidate" on the same page does:

- Save: no \`ConfirmButton\` (\`app/admin/names/page.tsx:107-125\`)
- Invalidate: wrapped in \`ConfirmButton\` (\`app/admin/names/page.tsx:117-124\`)

The API itself does a straight overwrite with no history table:

\`\`\`ts
// app/api/admin/names/route.ts:60-66
await db.update(names).set({ data }).where(and(eq(names.slug, slug), eq(names.kind, kind)));
\`\`\`

## Impact
Editing cached AI-generated content for a Divine Name permanently overwrites the previous payload in a single, un-confirmed click — there's no way to see or revert to what was there before. A mistake in the JSON textarea (e.g. saving a half-edited draft, or content typed into the wrong field) is unrecoverable other than manually invalidating the entry and waiting for it to regenerate from scratch, which loses any manual admin edits that weren't AI-generated. This is inconsistent with how the Prompts page treats similar content — prompts are explicitly versioned with rollback support.

## Location
\`app/admin/names/page.tsx:107-125\`, \`app/api/admin/names/route.ts:60-66\`

## Suggested fix
At minimum, gate "Save" behind the same \`ConfirmButton\` pattern already used for "Invalidate" on the same page. Ideally, keep a lightweight history of prior \`data\` payloads per \`(slug, kind)\` (even just the previous version) so an accidental overwrite can be undone.

nazarli-shabnam:
## Summary

Closes #275.

On the admin names page, "Save" (`PATCH`es `/api/admin/names`) overwrites the stored `data` payload for a Divine Name outright, with no confirmation and no history to revert to — unlike "Invalidate" on the same page, which already confirms, and unlike Prompts, which is explicitly versioned with rollback support.

## Changes

- Gated "Save" behind the same two-click confirm pattern already used for "Invalidate" on this page.
- Implemented via `useArmedConfirm` directly (not the shared `ConfirmButton`) so the JSON textarea can be **locked while armed** — otherwise the confirm gate wouldn't actually bind to the content the admin saw when they clicked "Save" the first time (same class of gap caught in review on #274/PR #466 for the Prompts page's Create & activate button, applied proactively here).
- Extracted the editing row into its own `EditRow` component so the armed-confirm hook has per-row scope (can't call a hook conditionally inside `.map()`).
- Scoped to the confirmation gate only, per the issue's "at minimum" framing — did not add a history/versioning table, since none exists for names today and that's a larger feature better suited to a separate issue if wanted.

## AI / Theological changes

- [ ] N/A — this only gates when a save happens, not what content is saved or how it's generated. No prompt, divine-name data, or theological framing change.

## Test plan

- Added `__tests__/app/admin/names/page.test.tsx`: asserts the first click doesn't call the API, the textarea locks and the button relabels to "Save?" while armed, and only the confirming second click issues the `PATCH` with the edited content.
- `bun run lint`, `bun run typecheck`, `bun run test:ci` pass locally.
- `bun run test:integration` passed via the pre-push hook.

## Summary by CodeRabbit

* **New Features**
  * Added two-step save confirmation when editing names.
  * The first click arms the save action and changes the button to “Save?” while temporarily disabling the text field.
  * Cancel remains available before confirmation.

* **Tests**
  * Added coverage for the confirmation flow and verified updates are submitted only after the second click.

### DIFF
--- a/__tests__/app/admin/names/page.test.tsx
+++ b/__tests__/app/admin/names/page.test.tsx
@@ -1,0 +1,56 @@
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { describe, expect, it, vi, beforeEach } from "vitest";
+
+const mockApi = vi.fn();
+vi.mock("@/components/admin/AdminContext", async () => {
+  const actual = await vi.importActual("@/components/admin/AdminContext");
+  return {
+    ...actual,
+    useAdminFetch: () => mockApi,
+  };
+});
+
+import NamesPage from "@/app/admin/names/page";
+
+const row = {
+  slug: "ar-rahman",
+  kind: "pairings",
+  data: { note: "original" },
+  model: "claude-opus-4-7",
+  version: 1,
+  updatedAt: "2026-01-01T00:00:00.000Z",
+};
+
+describe("NamesPage — Save requires confirmation", () => {
+  beforeEach(() => {
+    mockApi.mockReset();
+    mockApi.mockResolvedValue({ rows: [row] });
+  });
+
+  it("does not overwrite on the first click, locks the textarea while armed, and saves the confirmed content on the second click", async () => {
+    render(<NamesPage />);
+
+    fireEvent.click(await screen.findByRole("button", { name: "Edit" }));
+
+    const textarea = await screen.findByDisplayValue(/"note": "original"/);
+    fireEvent.change(textarea, { target: { value: '{"note": "edited"}' } });
+
+    fireEvent.click(screen.getByRole("button", { name: "Save" }));
+
+    expect(mockApi).not.toHaveBeenCalledWith(
+      "/names",
+      expect.objectContaining({ method: "PATCH" })
+    );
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs, CLAUDE.md/AGENTS.md compliance, git history context, and prior PR feedback (including the confirmed-content mismatch class caught in review on PR #466 — verified that gap doesn't have a reachable equivalent here since EditRow unmounts entirely on row switch).

nazarli-shabnam:
@coderabbitai review please

nazarli-shabnam:
@coderabbitai anything found?

PR #468, opened 2026-08-13
fix(admin): guard override-winner against non-active/completed challe…

### LINKED ISSUE #270: fix: override-winner admin action has no status guard, unlike end
## Problem
In the admin challenge-moderation endpoint, the \`end\` action checks the challenge's current status before mutating it:

\`\`\`ts
if (body.action === "end") {
  if (challenge.status !== "active") {
    return NextResponse.json({ error: "Only active challenges can be ended" }, { status: 409 });
  }
  ...
}
\`\`\`

But the \`override-winner\` action has no equivalent check — it only validates that \`winnerId\` is a participant (or null), then unconditionally sets \`status: "completed"\`:

\`\`\`ts
if (body.action === "override-winner") {
  const winnerId = body.winnerId ?? null;
  if (winnerId !== null && winnerId !== challenge.challengerId && winnerId !== challenge.challengedId) {
    return NextResponse.json({ error: "winnerId must be a participant or null" }, { status: 400 });
  }
  const [updated] = await db
    .update(challenges)
    .set({ status: "completed", winnerId })
    .where(eq(challenges.id, challengeId))
    .returning();
  ...
}
\`\`\`

## Impact
A direct API call (or a future UI regression that renders the override control for more statuses than intended) can force a \`pending\`, \`declined\`, or \`cancelled\` challenge straight to \`completed\` with a fabricated winner. This corrupts the stats shown on the admin challenges page (\`byStatus\` counts, \`fromSuggestions\` totals) and produces a "won: @user" result for a match that was never played, declined, or was cancelled. The client UI currently only renders override buttons for \`active\`/\`completed\` rows, but that's UI-only protection — the API itself doesn't enforce it, unlike the \`end\` action right above it in the same file.

## Location
\`app/api/admin/challenges/[id]/route.ts:77-96\` (compare with the \`end\` action's guard at lines 42-45)

## Suggested fix
Reject \`override-winner\` unless \`challenge.status\` is \`"active"\` or \`"completed"\` (mirroring the \`end\` action's status guard), returning a 409 otherwise.

nazarli-shabnam:
The end action already checks challenge.status before mutating; the override-winner action had no equivalent check and could force a pending, declined, or cancelled challenge straight to completed with a fabricated winner, corrupting admin stats and producing a fake result for a match that was never played. Restricts override-winner to the same active/completed statuses the admin UI already scopes the control to, with the same optimistic-concurrency WHERE pattern already used by end.

## Summary

- Guard `override-winner` in `app/api/admin/challenges/[id]/route.ts` to only accept `active`/`completed` challenges (409 otherwise), matching the statuses the admin UI already renders the control for.
- Scope the update's WHERE to the checked status so a concurrent status change loses the race with a 409 instead of clobbering it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

## Checklist

- [x] Conventional commits

Closes #270

## Summary by CodeRabbit

* **New Features**
  * Admins can override winners for active and completed challenges.
* **Bug Fixes**
  * Winner overrides are now rejected for pending, declined, and cancelled challenges.
  * Conflicting updates are detected and return a conflict response, preventing accidental overwrites when challenge status or winner information changes concurrently.

### DIFF
--- a/__tests__/api/admin/challenges.test.ts
+++ b/__tests__/api/admin/challenges.test.ts
@@ -99,6 +99,30 @@ describe("admin challenges [id]", () => {
     expect(res.status).toBe(200);
   });
 
+  it("accepts override-winner on an already-completed challenge (correcting a mistaken result)", async () => {
+    mockSelect.mockReturnValue(makeDbChain([{ ...challenge, status: "completed" }]));
+    const res = await PATCH(
+      req("PATCH", { action: "override-winner", winnerId: challenge.challengerId }),
+      params
+    );
+    expect(res.status).toBe(200);
+  });
+
+  it.each(["pending", "declined", "cancelled"])(
+    "rejects override-winner on a %s challenge",
+    async (status) => {
+      mockSelect.mockReturnValue(makeDbChain([{ ...challenge, status }]));
+      const res = await PATCH(req("PATCH", { action: "override-winner", winnerId: null }), params);
+      expect(res.status).toBe(409);
+    }
+  );
+
+  it("409s when override-winner races a concurrent status change", async () => {
+    mockUpdate.mockReturnValue(makeDbChain([])); // scoped update matched nothing
+    const res = await PATCH(req("PATCH", { action: "override-winner", winnerId: null }), params);
+    expect(res.status).toBe(409);
+  });
+
   it("rejects an unknown action", async () => {
     expect((await PATCH(req("PATCH", { action: "nope" }), params)).status).toBe(400);
   });
--- a/__tests__/api/admin/challenges.test.ts
+++ b/__tests__/api/admin/challenges.test.ts
@@ -123,6 +123,18 @@ describe("admin challenges [id]", () => {
     expect(res.status).toBe(409);
   });
 
+  it("409s when two override-winner corrections race on an already-completed challenge", async () => {
+    // Status alone can't detect this race (it stays "completed" throughout),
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (fixed in 2e895b5):

1. The status-scoped concurrency guard on `override-winner` was a no-op for the completed→completed correction path. When the starting status is already `completed`, the update only changes `winnerId`, not `status`, so the `WHERE status = 'completed'` clause stays true before and after the write. Two concurrent corrections against an already-completed challenge would both match and the second would silently overwrite the first's `winnerId` with no 409 — contradicting the comment's own claim that a concurrent change "loses the race cleanly instead of silently clobbering it."

https://github.com/OpenHikmah/openhikmah-web/blob/87c99b11c1bfe7f8962f933e874d4c9c7d0b2de5/app/api/admin/challenges/%5Bid%5D/route.ts#L96-L111

Fixed by also pinning the previously-read `winnerId` in the completed branch's WHERE clause, and adding a regression test for the completed→completed race:

https://github.com/OpenHikmah/openhikmah-web/blob/2e895b588bfd2aec43325c82a7ff285452ebe053/app/api/admin/challenges/%5Bid%5D/route.ts#L98-L120

nazarli-shabnam:
@coderabbitai review please

PR #469, opened 2026-08-14
fix(admin): rate-limit self-heal writes in GET /api/admin/challenges

### LINKED ISSUE #271: fix: GET /api/admin/challenges performs self-heal DB writes with no rate limiting
## Problem
Every mutating admin route under \`app/api/admin/challenges/\` calls \`rateLimitAdminMutation(auth)\` before writing (see \`PATCH\`/\`DELETE\` in \`[id]/route.ts\`, and \`finalize/route.ts\`). But \`GET /api/admin/challenges\` self-heals ended/expired challenges as a side effect of a plain read — issuing real \`UPDATE\`s via \`resolveEndedChallenges\`/\`resolveExpiredPending\` — with only \`requireAdmin\`, no \`rateLimitAdminMutation\`:

\`\`\`ts
export async function GET(req: NextRequest) {
  const auth = await requireAdmin(req);
  if (auth instanceof NextResponse) return auth;
  // ... no rateLimitAdminMutation call ...
  const ended = await db.select().from(challenges).where(and(eq(challenges.status, "active"), lt(challenges.endsAt, now)));
  await resolveEndedChallenges(ended, now);
  const expiredPending = await db.select().from(challenges).where(and(eq(challenges.status, "pending"), lt(challenges.endsAt, now)));
  await resolveExpiredPending(expiredPending, now);
  ...
}
\`\`\`

## Impact
The admin challenges page calls this endpoint on every filter change and on mount, and it's a GET (proxies/browsers may prefetch or retry GETs, and it's easy to poll accidentally via a bookmarklet or script). Because it performs real writes with no throttle, it's the one write path in this route group that lacks the protection every other mutation in the same directory already has. Note: this is distinct from the already-fixed #176 (which was about the *cost* of self-healing being N+1 — now addressed here via \`mapWithConcurrency\`); this issue is specifically about the missing rate limit on the write path itself.

## Location
\`app/api/admin/challenges/route.ts:22-49\`

## Suggested fix
Apply \`rateLimitAdminMutation\` (or a lighter per-admin throttle) before the self-heal writes, or move the self-heal logic to a scheduled job / restrict it to the existing \`finalize\` endpoint so \`GET\` stays read-only.

nazarli-shabnam:
The self-heal writes triggered as a side effect of a plain GET (resolveEndedChallenges / resolveExpiredPending) were the one write path in `app/api/admin/challenges/` with no `rateLimitAdminMutation` guard, unlike every mutating handler in the same directory (`PATCH`/`DELETE` in `[id]/route.ts`, `finalize/route.ts`). The admin challenges page calls this endpoint on every filter change and mount, and it's a GET, so it's easy to hit repeatedly with no throttle.

## Summary

- `GET /api/admin/challenges` (`app/api/admin/challenges/route.ts`) now checks `rateLimitAdminMutation(auth)` before the self-heal writes, but only when there's actually something to heal (`ended.length > 0 || expiredPending.length > 0`) — plain reads with nothing to self-heal stay unaffected.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

## Checklist

- [x] Conventional commits

Closes #271

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved administrative challenge retrieval by applying throttling only when automatic data repairs are needed.
  * Prevented repair operations when a request is throttled.
  * Kept standard read-only requests available without unnecessary rate limiting.

* **Tests**
  * Added coverage for normal requests, repair scenarios, and throttled responses.

### DIFF
--- a/__tests__/api/admin/challenges-list.test.ts
+++ b/__tests__/api/admin/challenges-list.test.ts
@@ -1,0 +1,117 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest, NextResponse } from "next/server";
+import type { User } from "@/lib/infra/db/schema";
+
+const { mockRateLimitAdminMutation } = vi.hoisted(() => ({
+  mockRateLimitAdminMutation: vi.fn<() => Promise<NextResponse | null>>(async () => null),
+}));
+vi.mock("@/lib/admin/admin-auth", () => ({
+  requireAdmin: vi.fn(),
+  rateLimitAdminMutation: mockRateLimitAdminMutation,
+}));
+
+const { mockResolveEndedChallenges, mockResolveExpiredPending } = vi.hoisted(() => ({
+  mockResolveEndedChallenges: vi.fn(async () => new Map()),
+  mockResolveExpiredPending: vi.fn(async () => 0),
+}));
+vi.mock("@/lib/social/challenges", () => ({
+  scoreChallenge: vi.fn(async () => 0),
+  pickWinner: vi.fn(() => 1),
+  resolveEndedChallenges: mockResolveEndedChallenges,
+  resolveExpiredPending: mockResolveExpiredPending,
+  mapWithConcurrency: async (
+    items: unknown[],
+    _limit: number,
+    fn: (item: unknown) => Promise<void>
+  ) => {
+    for (const item of items) await fn(item);
+  },
+}));
+
+function makeSelectChain(resolveWith: unknown) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const chain: any = new Proxy(
+    function () {
+      return chain;
+    },
+    {
+      get(_t, prop) {
+        if (prop === "then")
+          return (res: (v: unknown) => unknown, rej?: (e: unknown) => unknown) =>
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (fixed in ddea368):

1. The rate-limit gate only guarded the first self-heal call. A second, independent write — `resolveEndedChallenges(rows)` on the list-enrichment path — ran unconditionally after the gate, using a fresh `new Date()` internally rather than the request's `now` snapshot. A challenge that flipped from active to overdue in the gap between the initial `ended` select and the later `rows` select would trigger a real UPDATE with no rate limiting at all, undermining the PR's own goal of rate-limiting every write path in this handler.

https://github.com/OpenHikmah/openhikmah-web/blob/26f274c17aa2c4c4a63b90f92b4b1f1a12b2b620/app/api/admin/challenges/route.ts#L80-L88

Fixed by detecting stale-active rows in the list query and gating that call the same way (with a consistent `now` passed through), plus a regression test:

https://github.com/OpenHikmah/openhikmah-web/blob/ddea368494ab7f74319205ffd3cbe85945925218/app/api/admin/challenges/route.ts#L80-L96

PR #470, opened 2026-08-14
fix(sidebar): check response status before removing a note optimistically

### LINKED ISSUE #262: fix: note delete in ContextSidebar updates UI optimistically without checking response status
## Problem
The note-delete handler removes the note from local state unconditionally after the fetch settles, without checking \`res.ok\`, and even a caught network error still falls through to the state update:

\`\`\`ts
const remove = async (id: number) => {
  if (!accessToken) return;
  await fetch(\`/api/notes/\${id}\`, { method: "DELETE", ... }).catch((e) => console.error(...));
  setNotes((prev) => prev.filter((n) => n.id !== id));
};
\`\`\`

## Impact
If the DELETE request fails server-side (4xx/5xx) or errors over the network, the UI still shows the note as deleted immediately, while it actually still exists in the database. The note will reappear the next time notes are reloaded, which is confusing — the user believed it was permanently removed and may have relied on that (e.g. re-typed sensitive content elsewhere assuming the original was gone).

## Location
\`components/layout/ContextSidebar.tsx:125-132\`

## Suggested fix
Only strip the note from local state when the response is \`ok\`; on failure, keep the note in state and surface an error (toast/inline message), mirroring the pattern already used for the \`save\` handler in the same file.

nazarli-shabnam:
`remove()` in `components/layout/ContextSidebar.tsx`'s `NotesSection` stripped a note from local state unconditionally after the DELETE fetch settled — even on a 4xx/5xx response or a caught network error — so a failed delete still looked successful in the UI until the notes list was reloaded.

## Summary

- `remove()` now only updates local state when the response is `ok`; on failure (non-ok response or thrown error) it keeps the note in state and surfaces an inline "Delete failed" error next to that note, auto-clearing after 3s — mirroring the existing `save()` handler's error pattern in the same file.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

## Checklist

- [x] Conventional commits

Closes #262

### DIFF
--- a/__tests__/components/layout/ContextSidebar.test.tsx
+++ b/__tests__/components/layout/ContextSidebar.test.tsx
@@ -67,8 +67,9 @@ describe("ContextSidebar — note delete checks response status", () => {
     vi.stubGlobal("fetch", mockFetch);
     mockFetch.mockReset();
     deleteResponse = async () => new Response(null, { status: 204 });
-    // Method-aware (rather than call-order-aware) so React 18's double-invoked
-    // effect in test mode doesn't consume the DELETE mock as the notes GET.
+    // Method-aware (rather than call-order-aware): sibling sections
+    // (InteractiveArabic's morphology fetch, etc.) also call this same mocked
+    // fetch, so a queued once-per-call mock can be consumed out of order.
     mockFetch.mockImplementation((_url: string, init?: RequestInit) =>
       init?.method === "DELETE"
         ? deleteResponse()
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (fixed in a0a38cd):

1. The new `deleteErrorId` timeout was never cleared on unmount, and wasn't scoped correctly across notes. `NotesSection` unmounts whenever the sidebar closes or a different verse is selected, so a pending `setTimeout(() => setDeleteErrorId(null), 3000)` could fire `setState` after unmount. Separately, since `deleteErrorId` is a single scalar, an earlier note's failed-delete timer could fire and clear a *different*, later-failing note's error message before its own 3s window elapsed. This is the same defect class already fixed twice elsewhere in this codebase via a `mountedRef` + timeout-ref pattern (`HikmahCanvas.tsx`, `CanvasToolbar.tsx`).

https://github.com/OpenHikmah/openhikmah-web/blob/f2162e389fd75c396860fadd2b8dab7ec70c5a06/components/layout/ContextSidebar.tsx#L126-L145

Fixed by applying the same established `mountedRef` + timeout-ref pattern (also fixing the identical pre-existing gap in the sibling `save()` handler), plus a regression test for the cross-note race:

https://github.com/OpenHikmah/openhikmah-web/blob/a0a38cdb8ced6d3fda2fc5b783bb2c3b0904ef2f/components/layout/ContextSidebar.tsx#L77-L168

nazarli-shabnam:
@coderabbitai review please

nazarli-shabnam:
@coderabbitai good catch — fixed in bf2e3ee: the ref is now cleared immediately before each new setTimeout assignment in both failure branches of remove(), and applied the same fix to save() for consistency. Added a regression test for the out-of-order-completion case (two deletes started concurrently, later-started one settles first).

PR #471, opened 2026-08-14
fix(admin): add useAsync keepDataOnError, opt JobRunner's poll into it

### LINKED ISSUE #272: fix: JobRunner's polling reload wipes the whole jobs table on a single transient error
## Problem
\`useAsync\` intentionally clears \`data\` to \`null\` whenever a reload errors, "so we never render error + old rows":

\`\`\`ts
} catch (err) {
  setError(err);
  setData(null); // never render error + old rows
} finally {
  setLoading(false);
}
\`\`\`

That's a reasonable default for a manual, user-triggered reload. But \`JobRunner\` calls \`reload()\` automatically every 4 seconds while a job is \`running\`, to show live progress:

\`\`\`ts
useEffect(() => {
  if (!isRunning) return;
  const id = setInterval(reload, 4000);
  return () => clearInterval(id);
}, [isRunning, reload]);
\`\`\`

## Impact
If one of those 4-second polls hits a transient error — an in-flight auth token refresh, a brief 5xx, a network hiccup — the entire jobs table disappears and is replaced by a bare error message. The admin loses visibility into the in-progress job's status/log until the *next* successful poll happens to fire and the table "flashes" back in. For a long-running seed/embedding job, this makes transient blips look like the job itself failed or the whole panel broke.

## Location
\`components/admin/useAsync.ts:39-42\`, \`components/admin/JobRunner.tsx:36-40\`

## Suggested fix
Give \`useAsync\` an option (e.g. \`keepDataOnError\`) to preserve the last-good \`data\` on background/polled reloads instead of clearing it, and have \`JobRunner\` opt into that mode for its interval-driven reloads while manual user-triggered reloads keep the current clear-on-error behavior.

nazarli-shabnam:
`useAsync` intentionally cleared `data` to `null` whenever a reload errored ("so we never render error + old rows") — a reasonable default for a manual, user-triggered reload. But `JobRunner` calls `reload()` automatically every 4 seconds while a job is running, to show live progress. If one of those polls hit a transient error (in-flight token refresh, brief 5xx, network hiccup), the whole jobs table disappeared and was replaced by a bare error message until the next successful poll happened to land.

## Summary

- `components/admin/useAsync.ts`: added a `keepDataOnError` option (default `false`, preserving existing behavior everywhere else) that preserves the last-good `data` on a failed reload instead of clearing it.
- `components/admin/JobRunner.tsx`: opts into `keepDataOnError: true` for its polling `useAsync` call, so a transient poll error surfaces the error banner without hiding the in-progress job's table/log.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

## Checklist

- [x] Conventional commits

Closes #272

## Summary by CodeRabbit

* **Bug Fixes**
  * Job Runner now keeps existing job information visible during temporary polling failures.
  * Failed reloads clear outdated data by default, while selected reloads can preserve the last successful results.
  * Improved handling prevents stale results from being retained after failed job actions.
  * Reload behavior is now more consistent when switching between data views.

* **Tests**
  * Added coverage for polling failures, failed reloads, data preservation, and stale-data clearing.

### DIFF
--- a/__tests__/components/admin/JobRunner.test.tsx
+++ b/__tests__/components/admin/JobRunner.test.tsx
@@ -1,0 +1,57 @@
+import { render, screen, waitFor, act } from "@testing-library/react";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+
+const mockApi = vi.fn();
+vi.mock("@/components/admin/AdminContext", async () => {
+  const actual = await vi.importActual("@/components/admin/AdminContext");
+  return {
+    ...actual,
+    useAdminFetch: () => mockApi,
+  };
+});
+
+import { JobRunner } from "@/components/admin/JobRunner";
+
+const runningJob = {
+  id: "seed-quran" as const,
+  label: "Seed Quran",
+  status: "running" as const,
+  startedAt: new Date().toISOString(),
+  completedAt: null,
+  error: null,
+  logTail: "working…",
+};
+
+describe("JobRunner — polling keeps the table on a transient poll error", () => {
+  beforeEach(() => {
+    mockApi.mockReset();
+    // shouldAdvanceTime lets the real clock keep the fake one ticking, so
+    // testing-library's setTimeout-based `waitFor` still resolves while the
+    // component's own `setInterval` (which must be fake for this test) runs.
+    vi.useFakeTimers({ shouldAdvanceTime: true });
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("keeps showing the last-good jobs table when a 4s poll errors", async () => {
+    mockApi.mockResolvedValueOnce({
+      jobs: [runningJob],
+      embedCoverage: { embedded: 10, total: 20 },
+    });
+
+    render(<JobRunner />);
+    await waitFor(() => expect(screen.getByText("Seed Quran")).toBeInTheDocument());
+
+    mockApi.mockRejectedValueOnce(new Error("transient 5xx"));
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(4000);
+    });
+
+    // The table (and the running job's row) must still be visible — a
+    // background poll blip shouldn't wipe it, only surface the error banner.
+    expect(screen.getByText("Seed Quran")).toBeInTheDocument();
+    expect(screen.getByText("Something went wrong")).toBeInTheDocument();
+  });
+});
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review

nazarli-shabnam:
@coderabbitai good catch — fixed in 7cba035: keepDataOnError is now a per-reload() option instead of a hook-instance option, so only the poll's own reload calls opt in. The action-triggered reload right after starting a job still clears on failure, so a failed post-action reload can't leave anyRunning stranded on pre-job data. Added a regression test.

nazarli-shabnam:
@coderabbitai I looked into this closely and I don't think the race is actually reachable, though the reasoning is subtle enough that I've locked it in as a permanent test (6e65ebb).

You're right that keepDataOnErrorRef is shared mutable state read at settle-time, not captured per-call. But every reload() call also bumps tick, and the data-fetching effect depends on [cacheKey, tick]. So when the poll's reload({ keepDataOnError: true }) fires while an earlier plain reload() is still in flight, React runs that earlier effect instance's cleanup (setting its own active = false) before the new effect starts — this happens on the render triggered by the tick bump, well before any real network fetch settles. When the superseded request's promise eventually resolves/rejects, its .catch checks `if (!active) return;` and bails immediately, never reaching the keepDataOnErrorRef.current check at all.

So a superseded request's outcome is discarded entirely — it can never apply a keepDataOnError value that wasn't in effect when it started, because by the time any other reload() could have changed the ref, that reload already invalidated this one via the same cancellation guard the effect already had before this PR. I verified this empirically (deferred promises, settling a superseded plain-reload after a poll-reload with a different keepDataOnError value already resolved) and it behaves as expected: data stays exactly as the poll left it, unaffected by the stale reload's later settlement.

Test: __tests__/components/admin/useAsync.test.tsx — "ignores a superseded reload's outcome entirely, even if a later reload changed keepDataOnError before it settles".

nazarli-shabnam:
@coderabbitai this one's valid — fixed in bda4928. keepDataOnErrorRef persisted across a plain cacheKey change (a different resource, driven by a prop change rather than reload()), so retained data from one key could leak into a failed fetch for a different key. No current caller can actually trigger it (JobRunner is the only keepDataOnError:true consumer and its cacheKey is a fixed literal), but useAsync is shared across 13 admin pages, so it's worth closing defensively rather than leaving a footgun for future consumers.

Reset the flag whenever cacheKey itself changes (tracked via a prevCacheKeyRef comparison at the top of the fetch effect), while still preserving whatever reload() set for same-key reloads. Added a regression test that fails without the fix and passes with it (verified by reverting the fix locally and confirming the new test catches it): __tests__/components/admin/useAsync.test.tsx — "doesn't let a retaining reload for one cacheKey leak into a later cacheKey's failed fetch".

PR #472, opened 2026-08-14
fix(nav): drop AccountMenu's incomplete ARIA menu semantics, add focus management

### LINKED ISSUE #265: fix: AccountMenu dropdown has inconsistent/incomplete ARIA menu semantics
## Problem
The \`AccountMenu\` dropdown is marked \`role="menu"\`, and its sign-out button has \`role="menuitem"\`, but the two \`Link\` items ("Saved canvases", "Friends & Leaderboard") have no \`role="menuitem"\`. There's no arrow-key navigation between items, opening the menu doesn't move focus into it (focus stays on the trigger button until Tab is pressed), and focus isn't returned to the trigger when the menu closes via outside-click or Escape.

## Impact
Screen-reader and keyboard users get an ARIA \`menu\` that behaves like a plain popover: some items are announced as menu items and some aren't (inconsistent semantics), there's no roving-tabindex/arrow-key support that the \`role="menu"\` pattern implies, and focus management doesn't follow the WAI-ARIA menu pattern. This is the same class of gap already tracked for canvas dialogs in #179 (ExpandMenu/CanvasTour focus trap), but here it's about menu role semantics specifically, on a different component (\`AccountMenu\`) not covered by that issue.

## Location
\`components/layout/AccountMenu.tsx:136-210\`

## Suggested fix
Either drop \`role="menu"\`/\`role="menuitem"\` entirely in favor of a plain nav popover with a simpler, correct a11y story (simplest fix), or implement the full WAI-ARIA menu pattern: consistent \`role="menuitem"\` on every item, arrow-key roving focus, initial focus moving into the menu on open, and focus returning to the trigger on close — following the same approach used for \`ExpandMenu\`/\`CanvasTour\`.

nazarli-shabnam:
`AccountMenu`'s dropdown was marked `role="menu"`, but only the sign-out button had `role="menuitem"` — the two `Link` items had none — and none of the roving-tabindex/arrow-key navigation a true ARIA menu implies was implemented. Opening the menu didn't move focus into it, and closing it didn't return focus to the trigger.

**Design choice, called out explicitly per the issue's own suggested options:** there's no existing WAI-ARIA menu implementation elsewhere in the codebase to extend (verified via grep — `AccountMenu.tsx` was the only file using `role="menu"`), so I went with the simpler, correct option: dropped `role="menu"`/`role="menuitem"` in favor of a plain nav popover (regular links/buttons, no artificial menu semantics to get wrong), rather than building out the full ARIA menu pattern (roving tabindex, arrow-key nav) from scratch for a single component.

## Summary

- Removed `role="menu"` from the dropdown panel and `role="menuitem"` from the sign-out button.
- Replaced `aria-haspopup="menu"` with `aria-controls` pointing at the panel, matching the WAI-ARIA APG's "popup button controlling non-menu content" pattern.
- Added standard popup-button focus management: focus moves to the first link when the panel opens, and back to the trigger button when it closes — however the close happens (Escape, outside click, or picking an item).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

## Checklist

- [x] Conventional commits

Closes #265

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved keyboard focus behavior when opening and closing the account popover.
  * Trapped and looped focus within the open popover.
  * Restored focus to the account trigger after the popover is closed.
  * Moved focus to the sign-in button after signing out.
  * Added accessible labeling relationships and removed misleading menu semantics.
* **Tests**
  * Added coverage for focus movement, authentication changes, and accessibility behavior.

### DIFF
--- a/__tests__/components/layout/AccountMenu.test.tsx
+++ b/__tests__/components/layout/AccountMenu.test.tsx
@@ -1,4 +1,4 @@
-import { screen, fireEvent, waitFor } from "@testing-library/react";
+import { screen, fireEvent, waitFor, act } from "@testing-library/react";
 import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
 import { renderWithIntl } from "../../test-utils/render-with-intl";
 
@@ -61,5 +61,32 @@ describe("AccountMenu — dropdown a11y", () => {
 
     fireEvent.keyDown(document, { key: "Escape" });
     await waitFor(() => expect(trigger).toHaveFocus());
+  });
+
+  it("traps Tab within the open panel, wrapping last to first and first to last", async () => {
+    renderWithIntl(<AccountMenu />);
+    fireEvent.click(screen.getByRole("button", { name: /account menu/i }));
+
+    const firstLink = await screen.findByRole("link", { name: /saved canvases/i });
+    const signOutButton = screen.getByRole("button", { name: /sign out/i });
+
+    await act(async () => {
+      signOutButton.focus();
+    });
+    fireEvent.keyDown(signOutButton, { key: "Tab" });
+    expect(document.activeElement).toBe(firstLink);
+
+    fireEvent.keyDown(firstLink, { key: "Tab", shiftKey: true });
+    expect(document.activeElement).toBe(signOutButton);
+  });
+
+  it("redirects focus to the Sign in button on sign-out, since the trigger unmounts", async () => {
+    renderWithIntl(<AccountMenu />);
+    fireEvent.click(screen.getByRole("button", { name: /account menu/i }));
+
+    const signOutButton = await screen.findByRole("button", { name: /sign out/i });
+    fireEvent.click(signOutButton);
+
+    await waitFor(() => expect(screen.getByRole("button", { name: /log in/i })).toHaveFocus());
   });
 });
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
### Code review

Found 2 issues (both fixed in bfab8d7):

1. The hand-rolled focus-management effect moved focus in on open and back to the trigger on close, but never trapped Tab — a keyboard user could tab straight out of the open panel into page content behind it. This is the same defect class the repo has already fixed twice via `@radix-ui/react-focus-scope` (`ExpandMenu`/`CanvasTour`, and the mobile `More` sheet), both explicitly established as the house pattern for exactly this "popover opens, focus moves in, Tab is trapped, focus returns to trigger on close" problem.

https://github.com/OpenHikmah/openhikmah-web/blob/f3679a2c614f2cf0d16b0a7f2d3a41dee4dc46e8/components/layout/AccountMenu.tsx#L59-L71

2. Signing out swaps the component to a completely different subtree (the logged-out "Sign in" button), unmounting the trigger button that the close-focus logic tries to refocus — leaving keyboard focus orphaned with no visible target.

https://github.com/OpenHikmah/openhikmah-web/blob/f3679a2c614f2cf0d16b0a7f2d3a41dee4dc46e8/components/layout/AccountMenu.tsx#L73-L82

Fixed by wrapping the panel in `FocusScope` (matching the established pattern, with `onMountAutoFocus`/`onUnmountAutoFocus` overrides since Radix's defaults exclude `<a>` elements from mount candidates and rely on a `document.activeElement` snapshot that isn't reliable across browsers), and redirecting focus to the persistent Sign in button on sign-out:

https://github.com/OpenHikmah/openhikmah-web/blob/bfab8d73d86760ef1e40cd3a04c1a7e6a94fd3be/components/layout/AccountMenu.tsx#L168-L188

PR #489, opened 2026-08-19
fix(admin): close job-runner start-guard race

### LINKED ISSUE #359: fix: job-runner's start guard has a check-then-await race, allowing two jobs to run concurrently
## Summary

`lib/admin/job-runner.ts`'s `startJob` is meant to enforce "one job at a time," but the `running` guard is checked synchronously and only *set* after an `await db.insert(...)` resolves — giving a window where two near-simultaneous job-start requests both pass the guard.

## Evidence

`lib/admin/job-runner.ts:67-92` (`startJob`):

```js
if (running) throw new Error(`Job "${running.jobId}" is already running`);
...
const [row] = await db.insert(jobRuns).values(...).returning(...);  // await here
...
running = state;   // only set AFTER the await
```

The caller (`app/api/admin/jobs/route.ts`) adds further await points (`requireAdmin`, `rateLimitAdminMutation`) before `startJob` is even invoked, widening the window. Two `POST /api/admin/jobs` requests fired close together (double-click, two admin tabs) can both pass the `if (running)` check before either sets it, both insert `job_runs` rows, and both `spawn()` a script.

Compare to `lib/names/name-content.ts`, whose `inFlight`/`reasonInFlight` maps correctly perform the check-and-set *synchronously* (no await in between) to avoid exactly this race.

The existing test only covers the sequential case — `__tests__/lib/admin/job-runner.test.ts:96-99` awaits one `startJob` call to fully resolve before calling it again, never firing two concurrent (un-awaited) calls, so the race is untested as well as unfixed.

## Impact

Two concurrent runs of the same job — e.g. two `embed-corpus.mjs` runs both writing to `verse_embeddings`, or two `seed-quran.mjs` runs — can execute simultaneously, defeating the module's own documented "one job runs at a time" invariant and risking corrupted/duplicated data or wasted compute.

## Suggested fix

Set `running` synchronously before any `await`, matching the pattern already used in `lib/names/name-content.ts`. If the DB insert subsequently fails, clear `running` in the catch path.

## Acceptance criteria

- Two near-simultaneous calls to `startJob` for the same/different jobs: only one succeeds, the second throws the "already running" error, before any DB write from the second call occurs.
- New unit test fires two un-awaited `startJob()` calls back-to-back and asserts exactly one succeeds.

nazarli-shabnam:
## Summary

- `startJob` in `lib/admin/job-runner.ts` is meant to enforce "one job at a time," but the `running` guard was checked synchronously and only *set* after an `await db.insert(...)` resolved — a window where two near-simultaneous `POST /api/admin/jobs` requests (double-click, two admin tabs) could both pass the guard and both spawn a script.
- Fix: claim the `running` slot synchronously immediately after the guard check, before any `await`, matching the check-and-set pattern already used in `lib/names/name-content.ts`'s `inFlight`/`reasonInFlight` maps. `runId` is filled in once the DB insert resolves; if the insert throws, the slot is cleared in the catch path so a failed start doesn't permanently wedge the guard.

**CODEOWNERS note:** this PR touches `lib/admin/` (gated per `AGENTS.md`'s Auth/security-surface guardrail). The change only tightens the existing concurrency guard — no change to auth or rate limiting.

## Type of change

- [x] Bug fix

## AI / Theological changes

## Testing

- [x] `bun run test:integration` passes locally (Testcontainers/Docker) — touches DB-backed admin logic

Added a test firing two un-awaited `startJob()` calls back-to-back for different jobs and asserting exactly one fulfills (the other rejects with "already running"), with `db.insert` called only once.

## Checklist

- [x] Conventional commits
- [x] `.env.example` updated if new environment variables were added
- [x] `README.md` updated if the feature changes the public interface

Closes #359

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate job starts when multiple start requests occur nearly simultaneously.
  * Ensured only one job run is created while subsequent attempts are correctly rejected as already running.
  * Improved cleanup when creating a job run fails.

### DIFF
--- a/__tests__/lib/admin/job-runner.test.ts
+++ b/__tests__/lib/admin/job-runner.test.ts
@@ -98,6 +98,18 @@ describe("startJob", () => {
     await expect(startJob("seed-quran", "qf-admin")).rejects.toThrow("already running");
   });
 
+  it("only lets one of two near-simultaneous (un-awaited) calls succeed", async () => {
+    const first = startJob("seed-morphology", "qf-admin");
+    const second = startJob("seed-quran", "qf-admin");
+    const results = await Promise.allSettled([first, second]);
+    const fulfilled = results.filter((r) => r.status === "fulfilled");
+    const rejected = results.filter((r) => r.status === "rejected");
+    expect(fulfilled).toHaveLength(1);
+    expect(rejected).toHaveLength(1);
+    expect((rejected[0] as PromiseRejectedResult).reason.message).toMatch(/already running/);
+    expect(mockInsert).toHaveBeenCalledTimes(1);
+  });
+
   it("clears the running guard once the child process closes", async () => {
     await startJob("seed-morphology", "qf-admin");
     lastChild.current?.emit("close", 0);
--- a/__tests__/lib/admin/job-runner.test.ts
+++ b/__tests__/lib/admin/job-runner.test.ts
@@ -110,6 +110,15 @@ describe("startJob", () => {
     expect(mockInsert).toHaveBeenCalledTimes(1);
   });
 
+  it("releases the running guard when the insert rejects, allowing a retry to succeed", async () => {
+    mockInsert.mockReturnValueOnce(makeDbChain(Promise.reject(new Error("insert failed"))));
+    await expect(startJob("seed-morphology", "qf-admin")).rejects.toThrow("insert failed");
+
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

nazarli-shabnam:
@coderabbitai review please

PR #93, opened 2026-06-21
admin: RBAC seam, audit log, curated VotD tables + wiring

nazarli-shabnam:
* **New Features**
  * Added support for curated verse of the day with optional reflections displayed on the today page.
  * Introduced admin audit logging to track administrative actions.
  * Enabled account disabling functionality for administrative use.
  * Added feature flags configuration system.

* **Tests**
  * Improved verse-of-day test isolation with better mocking.
  * Updated test fixtures across multiple test suites for consistency.

## New Features
* **Admin Panel** – Added a protected admin experience with dashboards and tools for auditing, AI usage stats, runtime feature flags, infrastructure checks/maintenance, and connection moderation.
* **Curated Verse of the Day** – Admins can create/edit day-by-day overrides (including reflections), shown on the Today card.
* **Account Disabling** – Admins can soft-disable and re-enable accounts.

## Improvements
* **Verse-of-Day Reliability** – Enhanced curated lookup with deterministic fallback to the standard selection.

## Tests
* Updated fixtures to include `disabledAt` and expanded coverage for curated VOTD behavior and admin auth.

PR #500, opened 2026-08-20
Feat/admin story flags

nazarli-shabnam:
## Summary by CodeRabbit

* **New Features**
  * Added an admin Stories management page for viewing, hiding, and restoring stories.
  * Added status, reason, and audit details for hidden stories.
  * Added story visibility controls so hidden stories no longer appear in listings, pages, or social previews.

* **Bug Fixes**
  * Prevented flagged stories from being accessible through direct links or metadata generation.

* **Tests**
  * Added comprehensive coverage for administration, visibility rules, flag updates, restoration, and error handling.


nazarli-shabnam:
@coderabbitai review very detailed please

nazarli-shabnam:
@coderabbitai why didnt u review?

nazarli-shabnam (commented):
@coderabbitai review please

nazarli-shabnam:
@coderabbitai review please?

nazarli-shabnam:
### Code review

Found 1 issue:

1. `reason` is written to the DB with no runtime type check, unlike `slug` and `hidden` on the same handler.

`app/api/admin/stories/route.ts`'s PATCH handler validates `slug` (`typeof slug !== "string"`) and `hidden` (`typeof hidden !== "boolean"`) before use, but `reason` is passed straight through to `flagStory(slug, reason ?? null, auth.user.qfId)` — and from there into a `text` column via `lib/stories/story-flags.ts` — with only a compile-time `{ reason?: string }` annotation on the parsed JSON body. A PATCH body like `{ slug: "x", hidden: true, reason: 12345 }` writes a non-string value straight into the DB and into the audit-log `meta`.

The sibling route `app/api/admin/names/route.ts` validates its analogous free-text `reason` field with `isNonEmptyString()`, and `app/api/admin/votd/route.ts` guards its `reflection` field the same way after an identical gap was flagged Critical in review on #93. AGENTS.md's "only validate at system boundaries" applies directly here — this PATCH body is a system boundary.

https://github.com/OpenHikmah/openhikmah-web/blob/9e9242972c59f1f9c513e868c1cbd83ef6bd6adb/app/api/admin/stories/route.ts#L54-L69

nazarli-shabnam:
Checked every comment on this PR against the current code before applying anything.

**Fixed** (d266425):
- PATCH body validation: a literal `null` body threw past the parse try/catch into an unhandled 500; now rejected with 400 before destructuring. `reason` now gets a runtime `typeof` guard (matches `app/api/admin/votd/route.ts`'s existing pattern for `reflection`) instead of being written to the DB/audit log unchecked — this was also flagged in the review comment above.
- Added an `aria-label` to the per-row reason input.
- `app/stories/[slug]/opengraph-image.tsx` now sets `dynamic = "force-dynamic"`, matching the sibling page (`app/stories/[slug]/page.tsx`) and `app/today/opengraph-image.tsx` — without it, a newly-flagged story's OG image could keep serving from a stale prerendered cache.
- New tests: PATCH rate-limit rejection, null-body rejection, non-string-reason handling, and an opengraph-image fallback-vs-story-card regression test.

**Not applied** — two CodeRabbit suggestions conflict with conventions already established elsewhere in this codebase:
- Removing the `// admin qfId` comment on `flaggedBy` and the one-line JSDoc on the route handlers / `story-flags.ts` functions: both are repeated patterns already in `lib/infra/db/schema.ts` (e.g. `createdBy`, `triggeredBy`) and every sibling admin route (`votd`, `names`, `prompts`) — removing them here would be inconsistent, not cleaner, and cuts against CodeRabbit's own Docstring Coverage warning on this same PR.
- Wrapping the flag mutation and the audit-log write in one DB transaction: `lib/admin/admin-audit.ts` documents `logAdminAction` as intentionally best-effort (swallows its own errors so a logging failure never blocks the action it records), and the one other route using a real transaction (`app/api/admin/prompts/route.ts`) deliberately calls `logAdminAction` after the transaction commits, not inside it.

PR #550, opened 2026-09-05
fix(layout): clear the fixed mobile nav bar on scrollable pages

nazarli-shabnam:
## Summary

On mobile, the landing page's "Saved" quick-link sat partially behind the fixed bottom `MobileNavBar` when scrolled to the end — the reported bug. Root cause: `PersonalHome.tsx`'s scroll container was missing both `min-h-0` (needed to actually scroll inside the flex-column ancestor chain in `app/page.tsx`) and bottom clearance for the nav bar. Its sibling `MarketingHero.tsx` already had both.

A broader audit found the same missing-bottom-padding pattern repeated on other routes that render `<MobileNavBar />`:

- `components/home/PersonalHome.tsx` — added `min-h-0` and `pb-[calc(72px+env(safe-area-inset-bottom))]` on mobile (the primary fix)
- `app/surah/[number]/page.tsx` — `min-h-screen` → `min-h-dvh`, same bottom clearance on the verse list so the last ayah isn't hidden
- `app/stories/[slug]/page.tsx` — `min-h-screen` → `min-h-dvh`, same bottom clearance on the footer
- `app/today/page.tsx` — same bottom clearance, defensively, for long reflection text on narrow viewports
- `app/canvas/CanvasPageClient.tsx` — `h-screen` → `h-dvh` (this shell combines a fixed height with `overflow-hidden` and has no scroll fallback, so on iOS Safari before the address bar collapses, `100vh` can under/over-shoot the true visible viewport)

All changes follow the existing correct pattern already used by `MarketingHero`, `/names`, `/search`, and `/stories` index pages.

## Test plan

- [x] New `e2e/mobile-nav-overlap.spec.ts` at a 390×844 viewport, covering all three previously-broken surfaces (landing page, surah reader, story detail) — verified to fail without the fixes (confirmed by temporarily reverting the `PersonalHome.tsx` change and re-running)
- [x] Full `bun run test:e2e` suite passes (47/47)
- [x] `bun run lint`, `bun run typecheck`, `bun run format:check` pass
- [x] `bun run test:integration` passes (pre-push hook)
- [ ] Manual check at a real mobile viewport recommended before merge (devtools 390×844 on `/`, `/surah/1`, `/stories/<slug>`, `/today`, `/canvas`)

No dependency changes, no auth/security-surface changes, no theological content or AI prompt changes — layout/CSS only.

## Bug Fixes

- Improved mobile layout sizing across the Canvas, story, Surah reader, Today, and home pages.
- Added mobile navigation spacing to prevent fixed navigation from obscuring content.
- Improved viewport handling and page scrolling on mobile devices, including safe-area support.
- Adjusted responsive spacing and footer/content positioning across supported screen sizes.

## Tests

- Updated mobile regression coverage to verify content remains clear of the navigation bar on key pages.

### DIFF
--- a/app/canvas/CanvasPageClient.tsx
+++ b/app/canvas/CanvasPageClient.tsx
@@ -140,7 +140,7 @@ export function CanvasPageClient() {
   }, []);
 
   return (
-    <div className="flex flex-col h-screen w-screen overflow-hidden bg-bg">
+    <div className="flex flex-col h-dvh w-full overflow-hidden bg-bg">
       <Suspense>
         <VerseLoader />
       </Suspense>
--- a/app/globals.css
+++ b/app/globals.css
@@ -66,6 +66,14 @@
   --shadow-floating: 0 16px 40px rgba(0, 0, 0, 0.5);
 }
 
+/* Bottom clearance for content that scrolls behind the fixed mobile
+   MobileNavBar (min-h-[58px] + safe-area-inset-bottom, rounded up with room
+   for its border/blur) — see components/layout/MobileNavBar.tsx. Kept in one
+   place so every page that renders the bar clears it by the same amount. */
+@utility pb-mobile-nav {
+  padding-bottom: calc(72px + env(safe-area-inset-bottom));
+}
+
 @layer base {
   *,
   *::before,
--- a/app/stories/[slug]/page.tsx
+++ b/app/stories/[slug]/page.tsx
@@ -54,7 +54,7 @@ export default async function StoryDetailPage({ params }: Props) {
   );
 
   return (
-    <div className="min-h-dvh bg-bg text-text-primary">
+    <div className="min-h-dvh bg-bg pb-mobile-nav text-text-primary md:pb-0">
       <LandingHeader />
       <MobileNavBar />
       <StoryActivityTracker slug={slug} />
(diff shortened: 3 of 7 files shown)


nazarli-shabnam:
Addressed code-review findings from an internal `pr-review-toolkit:code-reviewer` pass:

- **Shared `pb-mobile-nav` utility** (`app/globals.css`) replaces the repeated `pb-[calc(72px+env(safe-area-inset-bottom))]` across four files, so the clearance amount lives in one place.
- **Consistent convention**: moved bottom clearance to the root wrapper on `/surah/[number]`, `/stories/[slug]`, and `/today`, matching the pre-existing pattern already used by `/names`, `/search`, and `/stories` (index) instead of introducing a second one.
- **Removed dead `overflow-y-auto`** on `/today`'s `<main>` — it never actually scrolled (no `min-h-0`, non-fixed-height wrapper), so it was misleading and would have been a centered-overflow trap if the layout ever did become a real scroll container.
- **Fixed two of the three new e2e assertions**, which were passing even against the unfixed baseline:
  - `scrollIntoViewIfNeeded()` doesn't scroll to the true bottom — replaced with an explicit scroll to max `scrollTop`/`scrollHeight`.
  - The surah test measured the Arabic paragraph instead of the true last element (the translation paragraph below it).
  - The story test measured `<footer>`'s own box, which includes the sacrificial padding — replaced with a `Range`-based measurement of the actual rendered text.
  - Re-verified all three now fail against the true pre-fix baseline (confirmed by swapping in the `main` version of each page and rerunning).

Full e2e suite (47/47), lint, typecheck, format check, and integration tests all still pass.

PR #551, opened 2026-09-05
fix(names): add shared header and mobile nav bar to name detail page

nazarli-shabnam:
## Summary
- \`app/names/[slug]/page.tsx\` (e.g. \`/names/al-malik\`) rendered a bare custom header (Wordmark + a plain back-link) and never mounted \`MobileNavBar\` — unlike every sibling detail page (\`stories/[slug]\`, \`surah/[number]\`) and the \`/names\` index itself, all of which use the shared \`<LandingHeader />\` + \`<MobileNavBar />\` pair. On mobile this left the page with no primary navigation at all (only the in-page "All Names" link or the browser back button).
- Found via a manual Playwright pass over the public routes at 390×844 and desktop; confirmed the gap by comparing screenshots of \`/names\` and \`/stories/adam\` (both show the bottom tab bar) against \`/names/al-malik\` (nothing).
- This is the same nav-bar surface PR #550 (\`fix/mobile-nav-overlap\`) just hardened for \`/\`, \`/surah/[n]\`, and \`/stories/[slug]\` — this route was simply missed since it never had the bar to begin with.
- Brings the page in line with its siblings: swaps the custom header for \`<LandingHeader />\` + \`<MobileNavBar />\`, matches the \`min-h-dvh ... pb-mobile-nav ... md:pb-0\` wrapper, and moves the "All Names" back-link into the hero content block (styled like the surah page's "back to search" link).
- Added a regression case to \`e2e/mobile-nav-overlap.spec.ts\` covering this page, following the existing pattern (scroll to bottom, assert real content clears the fixed nav bar).

Not in scope: \`/bookmarks\`, \`/settings\`, and the \`AuthShell\`-based pages (\`/mentions\`, \`/social\`, \`/workspaces\`) also don't render \`MobileNavBar\`. Flagging rather than fixing — those are account/utility pages one level removed from the content-detail pattern (names/stories/surah), so it may be intentional. Worth a follow-up discussion if not.

## AI / Theological changes
N/A — no AI prompts, divine-name data, or theological framing touched; this is a layout/navigation parity fix only.

## Test plan
- [x] \`bun run typecheck\`
- [x] \`bun run lint\`
- [x] \`bun run test:ci\` (full unit suite, 1505 tests)
- [x] \`bun run test:e2e -- mobile-nav-overlap\` (4/4 passing, including the new case)
[...]

PR #552, opened 2026-09-05
feat(names): fall back to Gemini once when Claude generation is empty

nazarli-shabnam:
## Summary
- Divine-name reflection/pairings/verse-reason generation had no recovery when Claude failed — I hit this for real while testing PR #551, where pairings generation returned empty because of an Anthropic "credit balance too low" error. The only recovery path was "retry on the next request," with no automatic backup provider.
- `GEMINI_API_KEY` is already provisioned in every deployment (`.env.example`) and Gemini's free tier easily absorbs an occasional retry, so this adds a one-shot Claude → Gemini fallback: when the primary (Claude) result comes back empty, retry once on Gemini before giving up.
- Scoped entirely to `lib/names/name-content.ts` (a new private `resolveAndGenerate` helper used by both `getOrGenerateNameContent` and `getOrGenerateVerseReason`) — zero changes to `lib/ai/ai.ts`, the three `app/api/names/[slug]/*/route.ts` routes, or `lib/ai/translate.ts`, since they're already parametric over whatever `{provider, model}` pair they're handed.
- One-directional (Claude → Gemini only) and only triggers when the *primary* resolved provider is Claude — a deployment already configured for Gemini on this feature (`ai_provider_names=gemini`) is unaffected.
- Deliberately **not** applied to the "connections" verse-backfill feature, which is intentionally Gemini-only (AGENTS.md: "Gemini-only, free tier") with its own key-rotation resilience — this fallback never touches that code path.
- The persisted `model` column always reflects whichever provider actually produced the cached content (not just the originally-resolved one), preserving the existing attribution invariant documented on `resolveNamesModel`.
- A fallback attempt that itself throws degrades to the primary's (empty) result rather than propagating as an unhandled rejection — verified this matters via a real scenario in the existing `names-ai-validation` test suite (a translation-fallback path with no internal try/catch).
- Added a metrics counter (`names_ai_fallback_used`) so admin analytics can see how often Gemini rescues a failed Claude call.

## AI / Theological changes
N/A — no prompts, divine-name data, or theological framing touched; this only changes which provider generates existing prompts under a resiliency retry.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run test:ci` (full unit suite, 1509 tests, including 6 new/updated cases in `name-content.test.ts` covering: fallback success + correct persisted model, no-fallback when already on Gemini, and a throwing fallback degrading safely)
- [x] Hand-traced every test in `__tests__/api/names-ai-validation.test.ts` against this change (it mocks the provider to always "claude") — confirmed all pass unmodified, then verified by actually running that suite
- [x] Manual check against the local dev server: hit `/api/names/al-malik/reflection` and `/api/names/al-malik/pairings` — confirmed the fallback is attempted and degrades gracefully (200, empty content, no 500) end-to-end

## Summary by CodeRabbit

- **Bug Fixes**
  - Added a fallback to Gemini when Claude returns empty content while generating AI-powered names or verse explanations.
  - Successfully generated fallback content is now saved correctly.
  - If the fallback also fails, the original empty result is returned safely without an unhandled error.
  - Gemini is not retried when it is already the primary provider.

### DIFF
--- a/__tests__/lib/names/name-content.test.ts
+++ b/__tests__/lib/names/name-content.test.ts
@@ -168,8 +168,81 @@ describe("getOrGenerateNameContent", () => {
     const out = await getOrGenerateNameContent("al-malik", "verses", "en", 1, generate, isEmptyArr);
 
     expect(out).toEqual([]);
+    // Test env defaults to Claude (see vitest setup), so an empty result
+    // triggers one Gemini-fallback retry — also empty here, so still nothing
+    // to persist.
+    expect(generate).toHaveBeenCalledTimes(2);
+    expect(mockInsert).not.toHaveBeenCalled();
+  });
+
+  it("falls back to Gemini once when the primary (Claude) result is empty, and persists the Gemini result", async () => {
+    mockSelect.mockReturnValue(makeSelectChain([])); // miss
+    const ai = await import("@/lib/ai/ai");
+    const providerSpy = vi.spyOn(ai, "resolveProvider").mockResolvedValue("claude");
+    const modelSpy = vi
+      .spyOn(ai, "resolveModel")
+      .mockImplementation(async (_feature, provider) =>
+        provider === "claude" ? "claude-opus-5" : "gemini-3.7-flash"
+      );
+
+    const generate = vi.fn(async (resolved: { provider: string; model: string }) =>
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Issue #566, opened 2026-09-08
fix: AI-generation write paths lack the guardrails admin edits/translations have


nazarli-shabnam:
## Summary

AI-generation write paths are held to a lower correctness/theology bar than the admin editor and the translation paths. Extends #557, #558.

## Findings

**B1 — `lib/ai/connection-generator.ts:157-160` (med)**
`parseRawConnections`'s shape filter accepts `reason: ""` / whitespace (`typeof c.reason === "string"`). A blank-reason connection then passes `isValidRef` and is persisted — violating "every edge on the canvas links to that explanation". The translation paths (`translateCellReasons`, `app/api/names/[slug]/verses/route.ts:295`) reject blank; the English generation path does not.
- Fix: add `c.reason.trim() !== ""` to the filter. The existing `parsed.length > 0 && valid.length === 0` guard (`:165`) then throws `ConnectionParseError` → caller retries instead of caching a junk row.

**B2 — `lib/ai/graph-service.ts:100,138-163` (med) — extends #557**
Live non-`en` traffic on a cold cell runs a fresh **native-language** `generateConnectionsForCell(…, locale, …)` and persists native rows — contradicting the batch invariant "the verse SELECTION is never re-derived per locale" (`lib/ai/connection-batch.ts:28-33`). Later `findTranslationGaps` adds translated rows on top → a cell ends up with >3 mixed-provenance rows and non-`en` reasons that diverge from the admin-reviewed `en` canonical text.
- Fix: on a non-`en` miss, ensure/generate the `en` row first, then translate its reason (mirror `translateCellReasons`).

**B3 — `app/api/names/[slug]/reflection/route.ts` + `pairings/route.ts` → `lib/names/name-content.ts:190-206` (med, suspected)**
Any non-empty model reply is persisted verbatim as canonical, version-pinned "Believer's Reflection" / pairing text — no post-generation Tanzih or refusal screen, no review gate (only `s.trim() === ""` is rejected). A soft refusal or off-tradition paragraph slips through as canonical theology. Same shape as #557/#558 in an uncovered location.
- Fix: screen output for refusal markers and don't cache on match; or persist as unreviewed pending an admin review flag.

**B4 — `app/api/names/[slug]/pairings/route.ts:111-123` (low)**
Caches pairings with `name: ""` when the model's proposed name doesn't resolve to one of the 99; the admin editor (`app/api/admin/names/route.ts` `isValidPairings`) rejects `name: ""`.
[...]

PR #576, opened 2026-09-08
fix(ai): hold names generation + connection reasons to the same output bar as admin edits

nazarli-shabnam:
## What

Three AI write paths accepted model output that the admin editors / translation paths would reject. Partially addresses #566 (parts B1, B3, B4 — B2, the graph-service per-locale selection divergence, is a hot-path change handled on its own branch).

## Changes

**B1 — `lib/ai/connection-generator.ts`**
`parseRawConnections` kept entries with a blank / whitespace-only `reason` (`typeof "" === "string"` passed the shape filter). A connection with no explanation then persisted — "every edge on the canvas links to that explanation". Now requires `reason.trim() !== ""`. An array whose entries are all blank throws `ConnectionParseError` (caller retries); a partially-blank array keeps the good entries.

**B3 — refusal screen (new `lib/ai/refusal.ts`)**
`app/api/names/[slug]/reflection/route.ts` persisted *any* non-empty model reply verbatim as canonical, version-pinned theology — a soft refusal or "As an AI, I cannot…" disclaimer slipped through. New `looksLikeRefusal(text)` is deliberately narrow (anchored to the start, unambiguous openers only, so it never suppresses a genuine "The believer's realisation of…" paragraph). A match is treated like an empty result: logged, `incr("names_ai_refusal")`, not cached, retried next request. Applied to `pairings/route.ts` too for symmetry (pairings was already partly covered by its "no JSON array" guard).

**B4 — `app/api/names/[slug]/pairings/route.ts`**
Cached a pairing with `name: ""` when the model's proposed name didn't resolve to one of the 99 — the admin editor's `isValidPairings` rejects `name: ""`. Now drops the unresolved pairing (logged) before caching. All-unresolved → `[]` → not cached → retried.

## Tests

- `__tests__/lib/ai/refusal.test.ts` (new) — refusal openers vs. genuine theological prose.
- `__tests__/lib/ai/connection-generator.test.ts` — all-blank reasons → `ConnectionParseError`; partially-blank → keeps the good ones.
- `__tests__/api/names-ai-validation.test.ts` — reflection/pairings refusal → 200 + empty (not cached); unresolved pairing name → dropped. Updated one existing pairings case to use exact `DIVINE_NAMES` values so it exercises name resolution.

## Verification

`bun run typecheck` · `bun run lint` · affected suites — clean.

Refs #566

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/__tests__/api/names-ai-validation.test.ts
+++ b/__tests__/api/names-ai-validation.test.ts
@@ -104,9 +104,10 @@ describe("names AI routes — model output validation", () => {
     mockCallAI.mockResolvedValue(
       JSON.stringify([
         {
-          transliteration: "Ar-Rahim",
+          // Exact DIVINE_NAMES values so it resolves to one of the 99.
+          transliteration: "Ar-Rahīm",
           arabic: "الرَّحِيم",
-          explanation: "Balances majesty with mercy.",
+          explanation: "Balances universal grace with mercy specific to the believers.",
         },
         { transliteration: 42, arabic: null },
         "junk",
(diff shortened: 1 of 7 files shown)


nazarli-shabnam:
@coderabbitai review pls

PR #591, opened 2026-09-09
feat(admin): coordinate the running-job lock across processes (#517)

### LINKED ISSUE #517: job-runner: coordinate the running-job lock across app processes (pg_try_advisory_lock)
## Context

Follow-up from #516 (CodeRabbit review).

`lib/admin/job-runner.ts` guards "only one job at a time" with a module-level `running` variable. That is process-local: if the app is ever deployed with more than one Node process, two near-simultaneous `startJob` calls can each pass the guard, insert separate `job_runs` rows, and spawn duplicate child scripts.

For `backfill-connections` specifically that means **duplicated LLM spend** — the batch job pays per generation/translation call.

The current single-box deployment makes this latent, not active, so it was left out of #516's scope.

## Proposal

Acquire a Postgres advisory lock (`pg_try_advisory_lock` / session-scoped, or an `xact` lock plus a lease row) atomically before creating the `RunningJob` state and inserting the `job_runs` row. Release on child `close` / `error`. Requests on other processes that fail to acquire get the same "a job is already running" error the in-memory guard raises today.

## Acceptance

- Two concurrent `startJob` calls (simulated across connections) → only one child spawned, the other rejects.
- Lock is released on normal completion, failure, and spawn error.
- Integration test in `__tests__/integration/` exercising the contention path.

nazarli-shabnam:
Closes #517. Follow-up from the #516 CodeRabbit review.

## Problem

`lib/admin/job-runner.ts` guarded "only one job at a time" with a module-level `running` variable. That is process-local: run the app with more than one Node process and two near-simultaneous `startJob` calls can each pass the guard, insert separate `job_runs` rows, and spawn duplicate child work. For `backfill-connections` that means **duplicated per-call LLM spend**.

## Change

New `lib/admin/job-lock.ts` — a Postgres **session-level advisory lock** (`pg_try_advisory_lock` / `pg_advisory_unlock`) on a dedicated connection:

- **Acquired** in `startJob`, immediately *after* the synchronous in-memory `running` claim. The in-memory guard is kept and still decides same-process races — advisory locks are re-entrant within one session, so relying on the lock alone would let a second same-process caller through. The lock only arbitrates *across* processes.
- **Released** in `finishRun` (the single chokepoint every terminal path funnels through: spawn `close`/`error`, in-process success/`catch`), guarded by the same `runId` check as the in-memory slot so a double `close`+`error` from one child can't unbalance it. Also released on the insert-failure path.
- A `startJob` on another process that can't take the lock throws `"A job is already running"` — the same class of error the in-memory guard raises, which `app/api/admin/jobs/route.ts` already maps to a 400.

The lock connection is **outside the shared pool** (`max: 1`, `idle_timeout: 0`, `max_lifetime: null`) so a multi-hour loop-mode job never occupies one of the pool's ten slots and the connection isn't recycled out from under a running job. If the process dies, Postgres drops the session and releases the lock automatically.

Response codes unchanged (the issue mentions 409, but the route has no such mapping today and 400 is the existing contract — out of scope here).

## Tests

- `__tests__/lib/admin/job-runner.test.ts` — mocks `@/lib/admin/job-lock`; adds cases: another process holding the lock → `startJob` rejects and inserts no row (guard released for retry); lock released on insert failure; lock released once the child closes.
- `__tests__/integration/job-runner-lock.integration.test.ts` (new, Testcontainers) — a second real connection holds `pg_try_advisory_lock(JOB_LOCK_KEY)` → `startJob` rejects, no `job_runs` row; the lock is held for the run's duration and released on completion; the next job starts once it's free.

## Verification

- `bun run test:ci` — green (pre-commit)
- `bun run test:integration` — green (pre-push), includes the new contention test

## Note

Branched on #590 (playwright-core dedupe) so the pre-commit `typecheck` step passes locally; GitHub will retarget this PR to `main` once #590 merges.

### DIFF
--- a/__tests__/integration/job-runner-lock.integration.test.ts
+++ b/__tests__/integration/job-runner-lock.integration.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect, beforeEach, afterEach, afterAll, vi } from "vitest";
+import { EventEmitter } from "node:events";
+import postgres from "postgres";
+import { sql } from "drizzle-orm";
+import { db } from "@/lib/infra/db";
+import { JOB_LOCK_KEY } from "@/lib/admin/job-lock";
+
+// The runner spawns `bun scripts/<name>.mjs` for script jobs — irrelevant to the
+// lock behaviour under test, so stub it with an EventEmitter we can "close".
+class FakeChild extends EventEmitter {
+  stdout = new EventEmitter();
+  stderr = new EventEmitter();
+}
+const { mockSpawn, lastChild } = vi.hoisted(() => ({
+  mockSpawn: vi.fn(),
+  lastChild: { current: null as FakeChild | null },
+}));
+vi.mock("node:child_process", () => ({ spawn: mockSpawn, default: { spawn: mockSpawn } }));
+vi.mock("@/lib/ai/connection-batch", () => ({ runConnectionBatch: vi.fn() }));
+vi.mock("@/lib/ai/connection-batch-loop", () => ({ runConnectionBatchLoop: vi.fn() }));
+
+import { startJob } from "@/lib/admin/job-runner";
+
+// A second connection standing in for "another app process".
+const outsider = postgres(process.env.DATABASE_URL as string, { max: 1 });
+
+const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
+
+beforeEach(async () => {
+  mockSpawn.mockReset().mockImplementation(() => {
+    const child = new FakeChild();
+    lastChild.current = child;
+    return child;
+  });
+  await db.execute(sql`TRUNCATE job_runs RESTART IDENTITY`);
+});
+
+afterEach(async () => {
+  // Close out whatever the test started so the runner's in-memory guard and the
+  // advisory lock are both released before the next test.
+  lastChild.current?.emit("close", 0);
+  lastChild.current = null;
+  await sleep(20);
+});
+
+afterAll(async () => {
+  await outsider.end();
+});
+
+describe("job-runner advisory lock (integration, real Postgres)", () => {
+  it("refuses to start when another connection holds the lock, and writes no job_runs row", async () => {
+    const [{ locked }] = await outsider<{ locked: boolean }[]>`
+      select pg_try_advisory_lock(${JOB_LOCK_KEY}) as locked
+    `;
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review (self)

Reviewed for bug/concurrency correctness and AGENTS.md compliance. The core mutual-exclusion logic — synchronous in-memory claim → advisory lock → single guarded release in `finishRun` — checks out: acquire/release is exactly-once across every terminal path (spawn `close`, spawn `error`, `error`+`close` double-emit, in-process success/`catch`, insert-failure), no same-process or cross-process deadlock, and `max_lifetime: null` / `idle_timeout: 0` genuinely disable postgres.js's default 30–60 min connection recycler that would otherwise drop the session lock mid-job.

Two items found and fixed in `f21ce44`:

1. **`spawn()` throw could strand the lock.** A synchronous throw from `spawn` (bad args / resource exhaustion) after lock acquisition + row insert would leave `running` set and the advisory lock held with no `finishRun` scheduled — wedging the runner cross-process. Now wrapped: a throw routes through `finishRun` (records the failed run, releases lock + guard) and rethrows.

https://github.com/OpenHikmah/openhikmah-web/blob/f21ce4460763f70cda99a4cfe9d13badcff32636/lib/admin/job-runner.ts#L378-L406

2. **Docstring overstated crash-release + omitted session-lock limits.** Corrected: Postgres releases the lock only once it notices the dropped session (immediate on clean exit, ~keepalive lag on unclean kill), and the accepted limitations (long-idle connection vs PG restart/failover/`idle_session_timeout`; transaction poolers) are now spelled out with the lease-row + heartbeat fallback.

https://github.com/OpenHikmah/openhikmah-web/blob/f21ce4460763f70cda99a4cfe9d13badcff32636/lib/admin/job-lock.ts#L10-L28

Not changed: no keepalive/heartbeat on the lock connection. A session-scoped lock on a dedicated connection was the deliberate choice for the current single-box, direct-to-Postgres deploy (the alternative — a lease row + heartbeat — reintroduces the timer complexity across both execution paths). The residual "connection blip frees the lock while the job runs" exposure is now documented as a known limit with a clear upgrade path.

Verifies the #517 acceptance criteria: two concurrent `startJob` calls across connections → one child, the other rejects; lock released on completion / failure / spawn error; integration test in `__tests__/integration/`.

Issue #562, opened 2026-09-06
bookmarks/[ref] DELETE does no trim/normalization (POST does)


nazarli-shabnam:
`app/api/bookmarks/[ref]/route.ts:15-21` deletes on `verseRef = ref` verbatim — no `.trim()`, no `isValidRef` — unlike POST (`app/api/bookmarks/route.ts:41-44`). Scoped by `userId` so not an authz issue, but a ref stored with surrounding whitespace via an older path can't be deleted (POST trimmed it on write, DELETE doesn't on remove → key mismatch), producing an un-removable bookmark.

**Fix:** normalize identically on both sides.

Found during the PR #553 bug sweep.

Issue #557, opened 2026-09-06
translateCellReasons persists any non-empty model output as canonical localized theology


nazarli-shabnam:
The only guard on a translation result (`lib/ai/connection-batch.ts` `translateCellReasons`, and `name-content.ts`) is `translated.trim() === ""`. A reply like `"I can't help with that."`, `"Sure! Here is the translation: …"`, or an English echo is non-empty, so it's written into `connections.reason` / `name_verse_reasons.reason` and served as canonical localized content — cached ~forever.

**Fix direction:** before persisting, reject/normalize obvious label prefixes, reject output byte-identical to the English source for a non-English locale, reject wild length-ratio mismatches. Log + skip (retry next run) rather than store.

Found during the PR #553 bug sweep.

PR #599, opened 2026-09-11
fix(names): retry on Gemini when the primary generation throws

### LINKED ISSUE #566: fix: AI-generation write paths lack the guardrails admin edits/translations have

nazarli-shabnam:
## Summary

- `lib/names/name-content.ts`'s `resolveAndGenerate()` retried the Claude→Gemini fallback only when the primary call resolved to an *empty* result — a primary call that **threw** propagated immediately with no retry, unlike every other write path in this area.
- Makes the two failure shapes symmetric: a throw now gets the same Gemini retry an empty result already gets. If the retry also fails, the **original primary error** is re-raised (not swallowed into a fabricated empty success) — only an empty-result primary still degrades to that empty result on a failed retry, since there's an actual valid value to fall back to in that case.

Closes #566. The issue's other three findings (B1 blank-reason connections, B2 native per-locale vs. en-canonical connections, B3 no refusal screening on reflection/pairings, B4 unresolved-name pairings cached) were already fixed by earlier PRs (#557 series, #562, #594). This ships the one remaining item, explicitly deferred in the issue's own comments pending this exact design ("assert: primary throw → Gemini retry attempted → original error re-raised only if retry also fails").

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

- [ ] No AI or theological changes in this PR
- [ ] Claude prompt modified — described below
- [ ] New/updated divine name descriptions — verified against Maturidi/Hanafi sources
- [ ] Changes to theological framing — described below
- [x] Contains AI-generated file(s)/block(s) with minimal human review — `Generated-By` trailer added per [AGENTS.md](../AGENTS.md#ai-attribution-policy)

**Details**: No prompt text or theological framing changed. This is purely retry/fallback plumbing around the existing "names" AI generation call (reflection/pairings/verses/verse-reason translation) — same prompts, same output validation, same Tanzih/refusal screening as before.

## Testing

## Checklist

- [x] Conventional commits
- [x] `.env.example` updated if new environment variables were added (n/a — none added)
- [x] `README.md` updated if the feature changes the public interface (n/a — internal retry behavior)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for name content and verse-reason generation when the primary provider encounters an error.
  - Generation now retries with a fallback provider when appropriate and preserves the fallback result if successful.
  - Clearer error handling ensures the original failure is surfaced when all retry attempts fail.
  - Prevented unnecessary retries when the configured provider is already the fallback provider.
  - Generation locks are cleared when retries cannot recover from an error.

### DIFF
--- a/__tests__/lib/names/name-content.test.ts
+++ b/__tests__/lib/names/name-content.test.ts
@@ -367,17 +367,69 @@ describe("getOrGenerateNameContent", () => {
     expect(out).toEqual(["ok"]);
   });
 
-  it("surfaces a generate() rejection and clears the lock so the next call retries", async () => {
+  it("retries on Gemini when the primary throws, and caches the fallback's result", async () => {
     mockSelect.mockReturnValue(makeSelectChain([])); // miss
-    const generate = vi.fn().mockRejectedValueOnce(new Error("boom")).mockResolvedValue(["ok"]);
+    const ai = await import("@/lib/ai/ai");
+    const providerSpy = vi.spyOn(ai, "resolveProvider").mockResolvedValue("claude");
+    const modelSpy = vi
+      .spyOn(ai, "resolveModel")
+      .mockImplementation(async (_feature, provider) =>
+        provider === "claude" ? "claude-opus-5" : "gemini-3.7-flash"
+      );
+
+    const generate = vi.fn(async (resolved: { provider: string; model: string }) => {
+      if (resolved.provider === "claude") throw new Error("claude overloaded");
+      return ["ok"];
+    });
+
+    const out = await getOrGenerateNameContent("al-malik", "verses", "en", 1, generate, isEmptyArr);
+
+    expect(out).toEqual(["ok"]);
+    expect(generate).toHaveBeenCalledTimes(2);
+    expect(mockInsert).toHaveBeenCalledTimes(1);
+    const values = mockValues.mock.calls[0][0] as Record<string, unknown>;
+    expect(values.model).toBe("gemini-3.7-flash");
+    providerSpy.mockRestore();
+    modelSpy.mockRestore();
+  });
+
+  it("does not retry on Gemini when a throwing primary is already Gemini", async () => {
+    mockSelect.mockReturnValue(makeSelectChain([])); // miss
+    const ai = await import("@/lib/ai/ai");
+    const providerSpy = vi.spyOn(ai, "resolveProvider").mockResolvedValue("gemini");
+    const modelSpy = vi.spyOn(ai, "resolveModel").mockResolvedValue("gemini-3.7-flash");
+    const generate = vi.fn().mockRejectedValue(new Error("gemini down"));
+
+    await expect(
+      getOrGenerateNameContent("al-malik", "verses", "en", 1, generate, isEmptyArr)
+    ).rejects.toThrow("gemini down");
+    expect(generate).toHaveBeenCalledTimes(1);
+    expect(mockInsert).not.toHaveBeenCalled();
+    providerSpy.mockRestore();
+    modelSpy.mockRestore();
+  });
+
+  it("surfaces the original error when the primary throws AND the Gemini retry also fails, and clears the lock so the next call retries", async () => {
+    mockSelect.mockReturnValue(makeSelectChain([])); // miss
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Self-reviewed with `pr-review-toolkit:code-reviewer` (CodeRabbit CLI isn't installed on this machine). Findings and what happened to each:

1. **The throw-retry path is effectively dead for AI-provider failures today** — every current `generate` callback (pairings, reflection, verses, translateReason) already catches its own `callAI` failure and returns empty rather than throwing, so a throw reaching `resolveAndGenerate` is never itself a provider error. A throw from the "verses" generator specifically comes from `searchVerseRefs`/`fetchVerseData` (non-AI network calls upstream of the LLM call), so retrying those against Gemini just re-runs the whole callback once more for an error a different provider can't fix, before correctly re-raising the original error.
   - **Accepted, documented, not reworked.** Rearchitecting `verses/route.ts`'s error handling to keep it out of this path is a larger, separate change with its own test-surface risk. Tightened `resolveAndGenerate`'s docstring instead to say plainly that this is dead work for AI failures today and exists as a uniform contract for any future `generate` that lets a real provider error propagate.
2. **A primary failure the Gemini retry papered over produced no log line or metric** — fixed. Now logs (`console.error`) and increments `names_primary_generation_failed` in the primary `catch` itself, before any retry is attempted.
3. **`primaryError === undefined` sentinel was unsafe** — `throw undefined` (or a falsy rejection) would've taken the success path and handed `isEmpty()` an actual `undefined`, throwing inside the predicate and masking the real failure. Fixed — replaced with an explicit `threw` boolean.
4. **Missing `getOrGenerateVerseReason` primary-throw test** (string-shaped `isEmpty`, not just the array-shaped case) — added.

All 43 tests in `name-content.test.ts` and the full 1615-test suite pass; `typecheck`/`lint`/`format:check` clean.

Issue #570, opened 2026-09-08
chore: plan nonce-based CSP enforcement — current policy would break Next hydration + GA


nazarli-shabnam:
## Summary

**Design task — must be planned before code.** #125 ("move CSP from report-only to enforced") is currently a trap: flipping the header name from `Content-Security-Policy-Report-Only` to `Content-Security-Policy` with the policy as written in `next.config.ts` would break production.

## Evidence (from prod browser console, openhikmah.com)

The current report-only policy logs these violations on a normal page load:

1. **`script-src 'self'` blocks Next.js's own inline hydration/streaming scripts** — many `self.__next_f.push(...)` inline `<script>` blocks (plus one inside a `_next/static/chunks/*.js`). With no `'unsafe-inline'`, nonce, or hash, enforcing this **stops the app from hydrating**.
2. **`script-src 'self'` blocks Google Analytics** — `https://www.googletagmanager.com/gtag/js` (loaded via `@next/third-parties`), not in any allowlist.
3. **`connect-src 'self'` blocks GA beacons** — `analytics.google.com`, `stats.g.doubleclick.net`.
4. **`img-src 'self' data: blob:` blocks the GA pixel** — `google.<tld>/ads/ga-audiences`.
5. `style-src` already carries `'unsafe-inline'` (the config comment covers this one).

## Work

1. Decide GA's fate: keep it (needs `script-src`/`connect-src`/`img-src` allowlist entries for the Google hosts) or drop it.
2. Adopt a **nonce-based CSP** — Next 16 supports per-request nonces via middleware/`headers()`; emit a nonce and add `'nonce-<value>' 'strict-dynamic'` to `script-src`. Verify OAuth (`form-action https://*.quran.foundation`) and the @xyflow canvas still work.
3. Roll out: report-only with the new policy → observe `/api/csp-report` clean for a window → flip to enforced. Then close #125.
4. Also fold in the `/api/csp-report` rate-limit/sampling (tracked in the API-hardening issue) so the reporting endpoint can't be used as a log-flood vector during the observation window.

## Related
- #125 (this unblocks it), `next.config.ts:37-66`, `proxy.ts`.

nazarli-shabnam:
Already resolved by #606 (merged) — added per-request CSP nonce infrastructure (report-only). This unblocks #125 (flip CSP from report-only to enforced) to proceed using the nonce plan designed here. Closing this design/planning issue; tracking the actual enforcement flip in #125.

Issue #125, opened 2026-07-03
Move CSP from report-only to enforced


nazarli-shabnam:
## Problem
`next.config.ts` (line 30) sets the `Content-Security-Policy-Report-Only` header rather than an enforced `Content-Security-Policy`. Per an inline comment, this is intentional for now — blocked on remaining inline styles in the app and the lack of a CSP violation-report endpoint.

## Proposed fix
- Add a report endpoint (e.g. `app/api/csp-report/route.ts`) to collect violation reports.
- Audit and remove/refactor remaining inline `style` usage (or move to nonce/hash-based allowances) that would otherwise be blocked once enforced.
- Flip the header from `Content-Security-Policy-Report-Only` to `Content-Security-Policy` once the above is clean, monitoring the report endpoint for a period before fully cutting over.

## Why this matters
Report-only CSP provides visibility but no actual protection against injected scripts/styles — it's a known, tracked gap rather than an oversight, but should be closed.

PR #606, opened 2026-09-12
feat(security): add per-request CSP nonce infrastructure (report-only)

nazarli-shabnam:
## Summary

- Issue #570: flipping CSP from report-only to enforced with the policy as written would break production (blocks Next's own inline hydration scripts, Google Analytics, and GA's beacon/pixel hosts — see the issue for prod console evidence).
- This PR ships **nonce infrastructure only** — the header stays `Content-Security-Policy-Report-Only`. Enforcing, closing #125, and `/api/csp-report` rate limiting are explicit follow-ups that need a real production observation window.
- `proxy.ts` generates a fresh per-request nonce, exposes it via an `x-nonce` request header, and sets a nonce'd `script-src 'self' 'nonce-<value>' 'strict-dynamic' https://www.googletagmanager.com` (overwriting `next.config.ts`'s static fallback for every route the proxy matches).
- `app/layout.tsx` reads the nonce via `headers()` and passes it to `<GoogleAnalytics nonce={...}>`.
- `connect-src`/`img-src` allowlist the specific GA hosts observed violating in prod (`analytics.google.com`, `stats.g.doubleclick.net`, best-effort `https://*.google.com` for the regional-TLD ad-audience pixel — CSP host-source syntax can't wildcard across TLDs, so occasional remaining reports there during the observation window are a known GA limitation, not a bug in this change).
- **Google Analytics is kept** (per product decision) — this PR adds the allowlisting/nonce plumbing GA needs rather than removing it.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run test:ci` passes locally (1619 tests)
- [x] `bun run format:check` passes locally (this file only — pre-existing warnings on unrelated `.superpowers/` files)

`__tests__/proxy.test.ts` adds: nonce present and well-formed in `script-src`, a fresh nonce per request, the nonce reaching the `x-nonce` request header (via Next's own `x-middleware-request-*` convention, verified against `node_modules/next/dist/server/web/spec-extension/response.js`), and the nonce'd CSP still being set on the maintenance-mode 503 response.

**Not done**: a live `bun run dev` + browser devtools console check (this machine hit memory constraints running a dev server alongside everything else this session) — worth a manual pass before merge to visually confirm zero *new* report-only violations vs. today's baseline.

## Auth/security surface note

This touches `proxy.ts` and CSP/security headers — not literally `lib/auth/`, `app/callback/`, or `app/api/admin/`, but security-adjacent per AGENTS.md guardrails, so flagging explicitly: the change does **not** touch the maintenance-mode gating logic or its matcher (unchanged), only adds nonce generation and a CSP header alongside it. The existing `form-action 'self' https://*.quran.foundation` OAuth directive is untouched.

## Checklist

- [x] Conventional commits
- [x] `README.md` updated if the feature changes the public interface — N/A, internal security header only

Related to #570 (does not close it — enforcement flip is a follow-up)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SRprUveCywFmhoTuFzLBGH

## Summary by CodeRabbit

- **Security**
  - Added per-request Content Security Policy reporting with unique nonces.
  - Applied CSP headers to regular responses and maintenance-mode responses.
  - Forwarded nonce and CSP information so supported scripts can load securely.
  - Added Google Analytics connectivity and reporting directives to the policy.
  - Development environments allow the additional scripting capability required for local operation.

- **Tests**
  - Added coverage for nonce generation, header forwarding, CSP directives, and maintenance responses.

### DIFF
--- a/__tests__/proxy.test.ts
+++ b/__tests__/proxy.test.ts
@@ -37,6 +37,46 @@ describe("proxy (maintenance mode)", () => {
   });
 });
 
+describe("proxy (CSP nonce)", () => {
+  beforeEach(() => {
+    mockGetFlagBoolean.mockResolvedValue(false);
+  });
+
+  it("sets a nonce'd Content-Security-Policy-Report-Only response header", async () => {
+    const res = await proxy(req("/"));
+    const csp = res.headers.get("Content-Security-Policy-Report-Only");
+    expect(csp).toMatch(/script-src 'self' 'nonce-[A-Za-z0-9+/=]+' 'strict-dynamic'/);
+    expect(csp).toContain("https://www.googletagmanager.com");
+    expect(csp).toContain("https://analytics.google.com");
+    expect(csp).toContain("report-uri /api/csp-report");
+  });
+
+  it("uses a fresh nonce on every request", async () => {
+    const csp1 = (await proxy(req("/"))).headers.get("Content-Security-Policy-Report-Only");
+    const csp2 = (await proxy(req("/"))).headers.get("Content-Security-Policy-Report-Only");
+    const nonce = (csp: string | null) => csp?.match(/'nonce-([A-Za-z0-9+/=]+)'/)?.[1];
+    expect(nonce(csp1)).toBeTruthy();
+    expect(nonce(csp1)).not.toBe(nonce(csp2));
+  });
+
+  it("exposes the same nonce on the x-nonce request header forwarded upstream", async () => {
+    const res = await proxy(req("/"));
+    const csp = res.headers.get("Content-Security-Policy-Report-Only");
+    const nonce = csp?.match(/'nonce-([A-Za-z0-9+/=]+)'/)?.[1];
+    // NextResponse.next({ request: { headers } }) surfaces the forwarded
+    // request headers on this special header (Next's own convention) so
+    // tests can assert what would reach the Server Component via headers().
+    expect(res.headers.get("x-middleware-request-x-nonce")).toBe(nonce);
+  });
+
+  it("still sets the nonce'd CSP header on the maintenance-mode 503 response", async () => {
+    mockGetFlagBoolean.mockResolvedValue(true);
+    const res = await proxy(req("/"));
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review

Found 2 issues (both fixed in 0fde22c415c19a2f9c9b16f6da484694481acc3e):

1. The nonce'd CSP was only set on the *response* headers. Next's own renderer reads the CSP off the incoming *request* headers to auto-extract the nonce and apply it to framework-generated inline scripts (hydration/flight-data payloads) — a response-only header left that auto-injection silently unwired, which would only surface once script-src is later flipped to enforced.

https://github.com/OpenHikmah/openhikmah-web/blob/1bbe9bb37f790559040683ea7d69e3dc7a419398/proxy.ts#L68-L74

2. `img-src`'s `https://*.google.com` was meant to cover GA's regional-TLD ad-audience pixel (`google.<tld>/ads/ga-audiences`), but CSP host-source syntax can only wildcard subdomains, never TLDs — it didn't solve the stated problem while needlessly widening img-src to every other google.com subdomain.

https://github.com/OpenHikmah/openhikmah-web/blob/1bbe9bb37f790559040683ea7d69e3dc7a419398/proxy.ts#L30-L33

Also verified (not a bug, but was flagged during review as an unsupported assumption): whether Proxy's `.set()` on `Content-Security-Policy-Report-Only` genuinely replaces next.config.ts's static header for the same route, rather than duplicating it. Confirmed directly against the installed Next 16.2.10 source (`node_modules/next/dist/server/lib/router-utils/resolve-routes.js`): the `fsChecker.headers` route runs before the `middleware` route in the `routes` array, and both write into the same `resHeaders` object via plain `resHeaders[key] = value` assignment — so it replaces, not duplicates. Cited in the updated code comment.

PR #610, opened 2026-09-12
fix(search): match quran.com keyword search against the user's UI locale

nazarli-shabnam:
## What
\`keywordSearch()\` in \`app/api/search/route.ts\` hardcoded \`language=en\` when calling quran.com's full-text search API, so ayah keyword search always matched the English corpus regardless of the caller's \`oh_locale\` cookie — even though the surah-name-match path and verse hydration were already locale-aware.

## Fix
- Added \`QURAN_API_LANGUAGE_BY_LOCALE\` in \`lib/i18n/config.ts\` mapping each of our locales (\`en/tr/ru/az\`) to quran.com's \`language\` query param.
- \`keywordSearch()\` now takes the caller's \`uiLocale\` and uses the mapped language instead of a hardcoded \`en\`.
- Verified against the live quran.com \`/search\` endpoint that all four language codes return real hits (Turkish "merhamet", Russian "милосердие", Azerbaijani "rəhmət" all matched correctly).
- Hydration (\`hydrate()\`) already used the same \`edition\` cookie for display text, so matched and displayed text now stay consistent.

## Out of scope
Semantic ("related by meaning") search stays English-embedding-only by design (see its own code comment in \`lib/quran/semantic-search.ts\`) — a separate, larger effort, not addressed here per discussion with the reporting user.

## Testing
- Added unit tests in \`__tests__/api/search.test.ts\` covering the default \`en\` language and the \`tr\`/\`ru\`/\`az\` locale-to-language mapping.
- \`bun run test:ci\`, \`bun run lint\`, \`bun run typecheck\`, \`bun run format:check\` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Search requests now use the selected interface language when querying results.
  * English, Turkish, Russian, and Azerbaijani searches return results in the appropriate language instead of always defaulting to English.

### DIFF
--- a/__tests__/api/search.test.ts
+++ b/__tests__/api/search.test.ts
@@ -409,6 +409,32 @@ describe("GET /api/search", () => {
     expect(mockGetVerses).toHaveBeenCalledWith(["2:30"], "az.mammadaliyev");
   });
 
+  it("defaults to language=en for quran.com's keyword search when the UI locale is English", async () => {
+    mockFetch.mockResolvedValueOnce(quranComResponse([]));
+    await GET(makeSearchReq("mercy"));
+    expect(mockFetch).toHaveBeenCalledWith(
+      expect.stringContaining("language=en"),
+      expect.anything()
+    );
+  });
+
+  it.each([
+    ["tr", "merhamet"],
+    ["ru", "милосердие"],
+    ["az", "rəhmət"],
+  ])(
+    "passes the caller's UI locale (%s) as quran.com's keyword-search language",
+    async (locale, q) => {
+      mockGetUiLocale.mockResolvedValue(locale);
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
Self-review (code-reviewer agent) surfaced two real issues, both fixed in f9bfecb:
- A non-English keyword search that genuinely found zero hits (rather than failing) now retries once in `en`, so an English/transliterated query someone types under a non-English locale still works — this was a real regression vs. pre-PR behavior.
- Dropped an unreachable `?? "en"` fallback (the locale map is total and `getUiLocale()` already whitelist-validates), fixed a stale comment, and softened an unverified-sounding claim in a doc comment.

Added 3 new tests covering the retry-on-zero-hits path, the no-retry-on-upstream-failure path, and end-to-end hydration under a non-English locale.

Issue #625, opened 2026-09-18
Give /admin a nonce'd CSP so it can be enforced too


nazarli-shabnam:
## Problem
proxy.ts's matcher deliberately excludes `admin`, `api/admin`, `api/auth`, `api/health`, `api/metrics`, and `api/csp-report` (for the maintenance-mode DB-flag escape hatch — an operator must always be able to reach the flag to turn maintenance back off, even if the DB read that backs it were somehow broken). Those routes fall back to `next.config.ts`'s static `Content-Security-Policy` header, which has no nonce (`script-src 'self'` only).

That's fine for `api/*` (JSON responses, no inline scripts) and static assets, but `/admin` is a real server-rendered React page. Confirmed via a production build (`next build` + standalone server) in #125's testing: enforcing CSP on that no-nonce fallback breaks `/admin` outright — Next's own framework-injected inline scripts (hydration/flight-data payloads) and the GTM script tag both get blocked, and the panel visibly errors out ("Connection closed").

## Proposed fix
Give `/admin` its own nonce, most likely by routing it through the proxy for nonce generation while keeping (or restructuring) the maintenance-mode short-circuit so the flag read still can't lock operators out. Once `/admin` has a nonce'd `script-src`, `next.config.ts`'s fallback CSP can move from `Content-Security-Policy-Report-Only` to enforced for the remaining routes (api/*, static assets) it still governs.

## Why this matters
`app/api/admin/` and the admin UI are called out in AGENTS.md as a high-risk surface. Its CSP currently offers no real script-injection protection (report-only), unlike the rest of the app after #125.

PR #626, opened 2026-09-18
fix(security): enforce nonce'd CSP on the main app surface

### LINKED ISSUE #125: Move CSP from report-only to enforced

nazarli-shabnam:
## Summary

- Flips `proxy.ts`'s per-request nonce'd CSP from `Content-Security-Policy-Report-Only` to enforced `Content-Security-Policy` for every route it matches (home, search, canvas, names, workspaces — the whole public app surface). Verified with a real production build (`next build` + the standalone server, not just dev) that no route in that set trips a CSP violation beyond one already-documented, accepted gap (a GA regional-TLD ad-audience pixel img-src violation, called out in the code comment).
- `next.config.ts`'s no-nonce fallback CSP (served only to `admin`, `api/*`, and static assets, which the proxy matcher excludes) **stays report-only**. Enforcing it broke `/admin` in the same production-build test: Next's own framework-injected inline scripts (hydration/flight-data payloads) and the GTM `<script>` tag both get blocked with no nonce to allow them, and the admin panel visibly errors out. Filed as a follow-up: #625.
- `style-src` keeps `'unsafe-inline'` in both — the canvas UI (`HikmahCanvas`, `HikmahEdge`, `ExpandMenu`, `CanvasPreview`, and the `names/[slug]` pages) sets inline `style={{...}}` for per-node/edge positions and colors computed at render time, values a static stylesheet can't express. `script-src` is the directive that actually stops injected-script XSS, and it's enforced with no `unsafe-inline`/`unsafe-eval` (dev-only `unsafe-eval` aside).

Closes #125

## Type of change

- [x] Bug fix (security hardening)
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run test:ci` passes locally (1697/1697)
- [x] New tests added: `__tests__/next-config-csp.test.ts` (regression guard keeping the no-nonce fallback report-only), `__tests__/proxy.test.ts` updated for the header-name change
- [x] Manually verified against a **production build** (`next build` + `node .next/standalone/server.js`) across `/`, `/search`, `/canvas`, `/names/[slug]`, and `/admin` with DevTools console open — this is how the admin-breakage finding above was caught; dev mode (`next dev`/Turbopack) alone showed a false-positive script-src violation from an HMR-only chunk that doesn't occur in production.

## Checklist

- [x] Conventional commits
- [x] `README.md` updated if the feature changes the public interface — N/A, no public-interface change

## Summary by CodeRabbit

- **Security**
  - Enforced Content Security Policy protection now applies to public application routes.
  - Nonce-based script controls help prevent unauthorized script execution.
  - Routes outside the proxy scope continue using a report-only policy for compatibility, including administrative pages and selected APIs and static assets.
  - Maintenance-mode error responses now include the enforced security policy.
  - Security policy behavior is covered by expanded regression tests.


nazarli-shabnam:
### Code review

Found 1 issue:

1. Both files' comments describe the proxy matcher's exclusions as "admin, api/*, and static assets," but the actual matcher only excludes a handful of specific `api/*` subpaths, not all of them — other API routes (e.g. `/api/search`) are matched and do get the newly-enforced nonce'd CSP. This PR rewrote both comments and kept the inaccurate shorthand, which could mislead someone reasoning about which API routes are/aren't covered by the enforcement this PR ships.

https://github.com/OpenHikmah/openhikmah-web/blob/30cd5378d5c41f431f1e809931699eda1b1e3b2d/proxy.ts#L111-L114

https://github.com/OpenHikmah/openhikmah-web/blob/30cd5378d5c41f431f1e809931699eda1b1e3b2d/next.config.ts#L34-L37

nazarli-shabnam:
Fixed in 177f19c: corrected the api/* matcher-exclusion description in both files, plus the stale `buildCspReportOnly` reference and dangling style-src cross-reference caught along the way.

nazarli-shabnam:
Verified against issue #125 and the actual diff — proxy matcher correctly excludes admin/auth/health/metrics/csp-report/static, the csp-report endpoint already exists, and the admin fallback staying report-only is the right call given the documented breakage. Tests updated correctly. Thank you for the thorough production-build verification!

PR #629, opened 2026-09-19
chore(quran): add --check mode to embed-corpus.mjs for coverage verification

### LINKED ISSUE #152: chore: verify corpus embedding backfill covers all 6,236 verses
## Problem
The project's internal roadmap explicitly flags this as unverified:

> Corpus embedding backfill (`scripts/embed-corpus.mjs`, Gemini) — status as of last update was "completing on paid tier"; verify before assuming it's finished.
>
> Verify embeddings cover all 6,236 verses post-backfill — effort XS

Semantic ("by meaning") search silently falls back to keyword search when embeddings are absent for a given verse, so a gap in embedding coverage wouldn't throw an error anywhere — it would just quietly degrade search quality for whichever verses are missing embeddings, with no signal to notice it happened.

## Proposed fix
- Run a query against the verse/embedding table to count verses with a non-null embedding vector vs. the total 6,236-verse corpus.
- If coverage is complete: close this issue.
- If coverage is incomplete: re-run `scripts/embed-corpus.mjs` (idempotent/resumable) for the missing verses, using `GEMINI_API_KEY`, then re-verify.
- Optionally, surface embedding coverage as a metric on `/api/metrics` (`lib/metrics.ts` already tracks `embed_cache_hit/miss` — a coverage percentage would be a natural addition) so this doesn't silently regress again in the future.

## Why this matters
This is explicitly called out as a verification gap in the project's own roadmap planning, and the fallback-to-keyword behavior means an incomplete backfill degrades silently rather than failing loudly — worth confirming directly rather than assuming it's done.

nazarli-shabnam:
## Summary

- Issue #152 asks to verify the corpus embedding backfill covers all 6,236 verses. No tool existed to check this without either running the full backfill (needs `GEMINI_API_KEY`) or a manual SQL query — this adds a `--check` mode to `scripts/embed-corpus.mjs`.
- `--check` counts `verses` total vs. `DISTINCT ref` in `verse_embeddings` for the current `EMBEDDING_MODEL`, prints the gap, and exits `1` if anything's missing / `0` if coverage is complete — no `GEMINI_API_KEY` needed for this mode, only `DATABASE_URL`.
- **Ran it against this machine's local dev Postgres**: `0/6236` embedded. This is expected, not a red flag — the local dev container is a separate, never-backfilled database from production.
- **I do not have production database access from this environment**, so I cannot verify actual production coverage myself. @nazarli-shabnam (or whoever has the production `DATABASE_URL`) — please run:
  ```
  DATABASE_URL=<production-url> node scripts/embed-corpus.mjs --check
  ```
  and report back. If it reports a gap, `node scripts/embed-corpus.mjs` (same env, no `--check`) is safe to re-run — it's idempotent/resumable, only fills the gap.
- Per the plan, deliberately **not** adding the "surface coverage on `/api/metrics`" enhancement the issue body suggests — it cites an `embed_cache_hit`/`embed_cache_miss` metric that doesn't actually exist in `lib/infra/metrics.ts` (confirmed via grep, repo-wide, case-insensitive). That's separate, unrequested scope; flagging it here as a possible follow-up rather than building it speculatively.
- This PR does not close #152 by itself — it delivers the verification tool; closing the issue depends on the production coverage number above.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run test:ci` passes locally (1696/1696; one run hit a transient `vitest-pool-runner` worker-timeout flake unrelated to this change, confirmed by a clean re-run)
- [ ] New tests added for new functionality — N/A; no existing test harness covers standalone `.mjs` scripts in this repo, and manual verification against a real Postgres (the log above) is a stronger check for this kind of DB-query script than a mocked unit test would be

## Checklist

- [x] Conventional commits
- [x] `README.md` updated if the feature changes the public interface — N/A

## Summary by CodeRabbit

* **New Features**
  * Added a check mode that reports embedding coverage without calling the embedding service.
  * Displays total verses, embedded references, and whether coverage is complete.
  * Returns a nonzero status when embedding coverage is incomplete.
* **Bug Fixes**
  * Coverage results now exclude embedding records that do not correspond to an existing verse.
* **Improvements**
  * Embedding credentials are no longer required when running coverage checks.


nazarli-shabnam:
### Code review

Verified the if/else restructuring preserves the original embed logic (resumable LEFT JOIN query, 429-retry/cap handling, upsert) unchanged in the else branch, `GEMINI_API_KEY` is correctly required only outside `--check` mode, and `process.exitCode` (not `process.exit()`) is used so the `finally` block's `sql.end()` still runs. No unresolved feedback from past PRs on this file applies here.

nazarli-shabnam:
Verified the --check logic against issue #152 — coverage math and the orphaned-embedding exclusion are correct, and no API key is required as claimed. Appreciate you being upfront that this doesn't close #152 by itself; I'll run the check against production and report back before closing the issue. Thanks!

Issue #563, opened 2026-09-06
Activity streak day is derived from a fully client-controlled tz offset


nazarli-shabnam:
`resolveActivityDate` (`lib/social/activity-date.ts`) treats `tz_offset_minutes` as authoritative. It's clamped to ±840 min but otherwise fully client-controlled, so a user can advance their streak day early / keep a broken streak alive by sending a favourable offset (e.g. at `2026-08-28T10:00:00Z`, `tz_offset_minutes: 840` resolves `local_date` to the 29th and can extend a streak whose previous day was the 28th).

Not a regression — the offset path predates PR #553 (that PR only tightened the *no-offset* fallback to bucket by UTC). Fixing it properly is a design change:

- persist a trusted user timezone (set once, changed rarely / rate-limited), and resolve the day from that; or
- accept the offset only within a small window of the server clock, and rate-limit offset changes.

Add an API test for the `tz_offset_minutes: 840` early-advance case once the approach is chosen.

From the CodeRabbit review on PR #553 (deferred — heavy lift, not a regression).

PR #630, opened 2026-09-19
fix(social): trust-anchor the client tz offset for streak days

nazarli-shabnam:
## Summary

- `app/api/social/activity` (`POST`) treated `tz_offset_minutes` as fully client-controlled, only sanity-clamped to ±840min. A user could send a favorable offset on every request to walk their "local today" boundary back and forth — advancing a streak day early, or resurrecting one that should have broken — without any real activity, exactly the exploit in #563.
- **Fix (trusted-anchor design, as agreed)**: a user's first-ever offset (or one within 60 minutes of their current anchor — covers DST) is trusted immediately and used to resolve the activity day. A bigger jump claims a genuine relocation and is rate-limited to **once per 20 hours** via the existing `lib/infra/rate-limit.ts` `consume()` — a faster repeat jump is rejected with `400` instead of silently moving the day boundary.
- No schema migration needed: the rate limit uses the existing `consume()` windowed-counter infra (Redis-primary, Postgres-fallback), not a new DB column, so `timezoneOffsetMinutes` continues to serve as both the stored anchor and the value readers (`GET /me`, leaderboard) decay against.
- The row lock (`FOR UPDATE`) already taken for the streak read/write now also gates the anchor comparison, closing a race where two concurrent requests could each see a stale anchor.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Documentation
- [ ] Tests

## AI / Theological changes

## Testing

- [x] `bun run test:ci` passes locally (1701/1701 — added 5 new tests for the anchor logic: first-offset trust, in-drift acceptance, rate-limited rejection of a big jump, successful jump once the limiter allows it, and that no DB write happens on rejection)
- [x] New tests added — see `__tests__/api/social/activity.test.ts`'s new "timezone anchor (issue #563)" block

## Checklist

- [x] Conventional commits
- [x] `README.md` updated if the feature changes the public interface — N/A, internal anti-abuse logic only

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved activity date and streak handling when a device’s timezone changes.
  - Small timezone shifts, including daylight-saving changes, are accepted automatically.
  - Large timezone changes are validated before updating activity records.
  - Requests may be rejected with a clear “Timezone changed too recently” message when changes occur too frequently.
  - Timezone updates are now saved even when the activity was already recorded for the day.


nazarli-shabnam:
Fixed in e53242b: moved the anchor-move rate-limit check (and its consume() I/O) out of the row-locking transaction, and corrected two now-stale docstrings in lib/social/streak.ts and lib/social/activity-date.ts that described the stored offset as "whatever the client last reported" instead of a vetted anchor.

nazarli-shabnam:
### Code review

Found and fixed 1 issue during self-review (already addressed in e53242b before this comment):

1. The anchor-move rate-limit check called `consume()` (Redis, or a Postgres write on its fallback path against the shared pool) while already holding a `FOR UPDATE` row lock on the user inside `db.transaction` — external I/O under an open lock and a second pooled connection risks pool contention under concurrent activity pings, a pattern this route never used before this PR.

https://github.com/OpenHikmah/openhikmah-web/blob/7063aa5f3b6e0e35bc3f76ba9c5b83c7ac0b13f4/app/api/social/activity/route.ts#L98-L124

Fixed by moving the anchor/trust decision (including `consume()`) to a plain, unlocked read before the transaction opens; the transaction still re-reads the row with `FOR UPDATE` for the existing streak-concurrency guarantee. Also corrected two now-stale docstrings (`lib/social/streak.ts`, `lib/social/activity-date.ts`) that described the stored offset as "whatever the client last reported" rather than a vetted anchor.

Verified: transaction restructuring preserves the atomic insert+streak-update guarantee and the row-lock-based concurrency fix from earlier history (`c1ba93d`), the `resolveActivityDate` UTC-fallback exploit fix from a prior bug sweep (`b53cb73`) is not reintroduced, and no other code path in the repo locks `users`/`activityLog`, so no new lock-ordering hazard.

nazarli-shabnam:
Verified against issue #563 and lib/infra/rate-limit.ts's consume() signature — the trusted-anchor design, the DST drift window, and the rate-limited-relocation path all check out, and the new tests cover the exploit case directly. Thank you for closing this cleanly!

